PR #4, opened 2022-02-15
Add per output screenshot.

Shinyzenith:
I don't have access to multiple monitors right now, if someone could test this then that would be great. 
I'm 99% sure this will work.


Shinyzenith:
Tested over HDMI, this works.

PR #5, opened 2022-02-17
Current monitor detection

Shinyzenith:


Shinyzenith:
Can someone with dual monitors test this?

Shinyzenith:
Also, fair warning, I do not know how to work with Option<T>, I kinda winged that part along with some help from rustc. If someone can review the logic then that would be amazing.

NikolaTesla13:
> Can someone with dual monitors test this?

I have two monitors, I'll test it later today when I get home.

Shinyzenith:
> > Can someone with dual monitors test this?
> 
> I have two monitors, I'll test it later today when I get home.

Thank you!

Shinyzenith:
checked rn, this does not work.

Shinyzenith:
XdgOutputHandler always returns is_current even if the monitor is not the one is focus, I guess my understanding of the parameter was flawed to begin with.

PR #27, opened 2022-07-28
[feat] set png compression/filter type through arguments

EdenQwQ:
Implemented the ability to set png compression/filter type with `-C` and `-F` argument
```rust
pub enum CompressionType {
    Default,
    Fast,
    Best,
    Huffman,
    Rle,
}

pub enum FilterType {
    NoFilter,
    Sub,
    Up,
    Avg,
    Paeth,
    Adaptive,
}
```


Shinyzenith:
Pr looks great! Is there anyway we can make this genetic enough for all the backend formats ?

EdenQwQ:
> Pr looks great! Is there anyway we can make this generic enough for all the backend formats ?

Compression type and filter type are options only available for png format. We can add options for other formats by using a similar smart pointer

Shinyzenith:
> > Pr looks great! Is there anyway we can make this generic enough for all the backend formats ?
> 
> Compression type and filter type are options only available for png format. We can add options for other formats by using a similar smart pointer

Ok I see, in that case, can we offload the compression to some other crate to do it for us which can support multiple formats?


If that's too big of a task right now, then it's just best to merge this for now :D

Also while we're on the topic, I added man-page generation in build.rs with scdoc, would you mind documenting this flag in the 1 and 7 sections of wayshot?

Shinyzenith:
This is a really good PR and I wish to get this to production someday. I just need to find a good way to compress all supported formats.

Gigas002:
Partially superseded by #268 (you can now set these options via config)

PR #31, opened 2022-09-30
Added 10-bit back buffer support

gigablaster:
I experiment with render_bit_depth option in sway and found that nothing can make screenshots in this mode. So I've added this feature. I only support Xbgr2101010 and Abgr2101010 because I don't have anything to test it on Xrgb2101010 or Argb101010 formats.

- Allow user to make screenshots if back buffer uses 10 bits per color channel.
- Image conversion moved to convert.rs

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -24,9 +24,7 @@ use image::{
         png::PngEncoder,
         pnm::{self, PnmEncoder},
     },
-    ColorType,
-    ColorType::Rgba8,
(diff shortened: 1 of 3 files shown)


Shinyzenith:
Super cool patch! Thanks for keeping everything properly commented and for the contribution.
I'll quickly go over it and if everything is fine, I'll merge it.

Shinyzenith (approved):

PR #32, opened 2022-12-04
[feature] add RGB565 buffer support

mstoeckl:
This makes it possible to take screenshots if the compositor renders to an RGB565 buffer.

I am submitting this PR because I already wrote the patch, and there's a chance it matches the project goals. However, as essentially no wlroots-based compositors currently allow rendering to RGB565 buffers, what with XBGR8888 or higher formats being preferable in almost all cases, I will understand and not be surprised if you close this.

I tested this with a [branch of sway](https://github.com/mstoeckl/sway/tree/extra-bit-depths) that lets you set the output bit format/depth with the command `swaymsg output '*' render_bit_depth 5`.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -38,7 +38,13 @@ use wayland_protocols::wlr::unstable::screencopy::v1::client::{
     zwlr_screencopy_manager_v1::ZwlrScreencopyManagerV1,
 };
 
-use crate::convert::create_converter;
+use crate::convert::{create_copy_converter, create_in_place_converter};
+
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


Shinyzenith:
Hi! Thanks for the patch. I'm all for new buffer supports and it does indeed align with the project goals.
However given that no compositor implements this yet, I'll hold back on merging. Also I'm not completely well versed in rgb565.
Note to compositor authors: if your compositor implements rgb565 buffers and you want this pr merged. A ping shall suffice!

Shinyzenith (approved):

PR #37, opened 2023-03-05
[update] Upgrade to wayland-rs 0.30.0

mstoeckl:
For this change, I am trying to stick as close to the current code as possible, and not adding any optimizations (like only creating a single wl_registry object, or using a single roundtrip for all xdg_outputs instead of one roundtrip per xdg_outputs, or reducing the amount of clone() being done).


Shinyzenith (approved):

Shinyzenith:
My apologies for the late review, due to medical conditions I've been set back severely on maintainence tasks.

Shinyzenith:
Speaking of the above mentioned optimisations, how would they be achieved? Would you mind telling me a bit about it? I'd like to try my hands at it.

Issue #25, opened 2022-07-15
Make wayshot a library/crate


wildwestrom:
I'm going to use like this anyway, but if you think this is a good idea, I can submit a PR for it.

Shinyzenith:
Actually I have been considering this, hence the backend.rs file is extensively documented and modular. I just never found the time to split it apart into a crate.

A patch for this is always welcome :D

PR #38, opened 2023-05-06
[feature] Split backend into libwayshot

Shinyzenith:
Closes: https://github.com/waycrate/wayshot/issues/25


Shinyzenith:
> > Do you think we should publish libwayshot as a crate with a minimal example in the repo?
> 
> Apologies for the late response. In general, I am not in favor of libraryification (or even just splitting things into many files), until there is a clear need for it: either someone who plans to use the library, or a point where the number of definitions grows large enough that a noticeable fraction of time is spent searching through a file for things. (Other people may have different preferences, but for me the point where I consider splitting files is usually around 1000 lines; it depends on the language, function sizes, and one's text editor, of course.)
> 
> With the above being said, now that this PR has been merged, I do not think that it should be reverted.

Hey! Thanks for taking a look, much appreciated ❤️. I totally get where you're coming from however I noticed a few downstream projects using wayshot source code by copying the entire `backend.rs` file which felt quite unintuitive. Taking note the downstream issue, this pr was opened. 

A quick example would be haruhishot (https://github.com/Decodetalkers/haruhishot). 
There has also been some interest about this earlier: https://github.com/waycrate/wayshot/issues/25

PR #41, opened 2023-06-27
[Feat] Multi-output composited screenshots

Shinyzenith:
I attempted to clean up the code a little bit and introduce some style changes to the same PR. The original author has been credited in the commit sign off.

@mstoeckl @Decodetalkers

Supercedes: https://github.com/waycrate/wayshot/pull/40

PR #44, opened 2023-07-02
[feat] Record output geometry transforms

mstoeckl:
This commit adds a field to `struct OutputInfo` corresponding to the output transform, and adjusts the code called by `get_all_outputs` to fill in this field using the [wl_output::geometry](https://docs.rs/wayland-client/latest/wayland_client/protocol/wl_output/enum.Event.html#variant.Geometry) event.

This is a small step toward supporting output transforms, and may be useful for https://github.com/waycrate/wayshot/pull/41 .


Shinyzenith (approved):

Shinyzenith:
Merged as: https://github.com/waycrate/wayshot/commit/4c931b1ebb19cc1f4a0ca1cc997f1dd1562a1bc6

PR #53, opened 2023-07-30
consolidate libwayshot api

### LINKED ISSUE #46: Libwayshot API inclusive of multi monitor setups.
Currently all multi output handling is done in wayshot itself. Maybe this could be consolidated into the libwayshot API itself ? 

I don't know if that would be a good API design but I presume most downstream projects would want to support multi-monitor setups and making it trivial in the API itself would reduce fragmentation in the implementations of multi-monitor and overall make it easier for the consumers of the API.

This thread is for further discussion on the topic, feel free to pitch in.

c-h-johnson:
Closes #46

I have not yet tested it with multiple outputs but everything else works.


Shinyzenith:
Would it be smart to wrap the display connection in the wayshot connection or should we rather have it exposed and be passed to the functions individually?

I propose this because the display can only be created in an application once and it might not just utilize wayshot, what if it is using xdg-shell to display that screenshot? In that case the app would need access to the display and the globals registry.

c-h-johnson:
`Connection` is `Clone` so the instance can be created using `WayshotConnection::from_connection(connection.clone())` but I don't know enough about wayland to know if it is a good idea. I don't think it can cause any problems.

Shinyzenith:
@Decodetalkers Would you mind testing this with multiple monitors once? I don't have access to the hardware at the moment.

libwayshot/src/screencopy.rs
```diff
@@ -0,0 +1,139 @@
...
+use anyhow::{anyhow, bail, Result};
+use image::{ColorType, DynamicImage, ImageBuffer, Pixel};
+use memmap2::MmapMut;
+use nix::{
+    fcntl,
+    sys::{memfd, mman, stat},
+    unistd,
+};
+use wayland_client::protocol::{
+    wl_output::{self},
```
Shinyzenith, line 16:
```suggestion
    wl_output,
```

Decodetalkers:
Yes,I will test it

libwayshot/src/lib.rs
```diff
@@ -93,383 +57,373 @@ pub struct CaptureRegion {
...
-impl From<EncodingFormat> for &str {
-    fn from(format: EncodingFormat) -> Self {
-        match format {
-            EncodingFormat::Jpg => "jpg",
-            EncodingFormat::Png => "png",
-            EncodingFormat::Ppm => "ppm",
-            EncodingFormat::Qoi => "qoi",
-        }
+impl WayshotConnection {
+    pub fn new() -> Result<Self> {
```
Shinyzenith, line 79: Maybe we should just keep from_connection. While having ::new would be convenient, it is more sensible to keep the API low level and provide more control.

libwayshot/src/lib.rs
```diff
@@ -93,383 +57,373 @@ pub struct CaptureRegion {
...
-impl From<EncodingFormat> for image::ImageOutputFormat {
-    fn from(format: EncodingFormat) -> Self {
-        match format {
-            EncodingFormat::Jpg => image::ImageFormat::Jpeg.into(),
-            EncodingFormat::Png => image::ImageFormat::Png.into(),
-            EncodingFormat::Ppm => image::ImageFormat::Pnm.into(),
-            EncodingFormat::Qoi => image::ImageFormat::Qoi.into(),
-        }
-    }
+/// # Example usage
```
Shinyzenith, line 66:
```suggestion
/// Struct to store wayland connection and globals list.
/// # Example usage
```

Shinyzenith:
Do we really need to add the anyhow crate? I'd rather not increase binary size. If does improve some aspects of the codebase then I'm totally willing to include it.

wayshot/src/wayshot.rs
```diff
@@ -99,192 +68,65 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     }
 
-    let capture_area = if let Some(slurpregion) = args.get_one::<String>("slurp") {
-        match utils::parse_geometry(slurpregion) {
-            Some(region) => (wl_output::Transform::Normal, region),
-            None => {
-                log::error!("Invalid geometry specification");
-                exit(1);
-            }
+    let image_buffer = if let Some(slurpregion) = args.get_one::<String>("slurp") {
```
Shinyzenith, line 86: It seems this might've missed my gaze during the last review. We use snake_case for variable names. Would you mind committing that?

wayshot/src/wayshot.rs
```diff
@@ -99,192 +68,65 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        if let Some(region) = utils::parse_geometry(slurpregion) {
+            wayshot_conn.screenshot((Transform::Normal, region), cursor_overlay)?
```
Shinyzenith, line 88: We should not need to hardcode the normal transform anymore. That was mostly a hack because we were going to set the transform later anyways. It would be more sensible to have that be abstracted away by the API.

If we *have* to hardcode normal just as a workaround ( which should no longer exist, ideally speaking ), then we should atleast mention that in the doc comments for the function.

libwayshot/src/lib.rs
```diff
@@ -93,383 +57,373 @@ pub struct CaptureRegion {
...
+            y_coordinate: y1,
+            width: x2 - x1,
+            height: y2 - y1,
+        };
+        self.screenshot((Transform::Normal, capture_region), cursor_overlay)
     }
 
-    Ok(())
+    /// Take a screenshot from all accessible outputs.
+    pub fn screenshot_all(&self, cursor_overlay: bool) -> Result<RgbaImage> {
```
Shinyzenith, line 426: Would `screenshot_all` composite all the output screenshots into one?

Shinyzenith:
Apart from a few minor questions and the transform::normal hack nitpick, the PR looks fine to me. Lets wait and see what Decode has to say about the multi-monitor behavior. If everything is fine, I'll do another indepth look one last time and merge.

Decodetalkers:
> Apart from a few minor questions and the transform::normal hack nitpick, the PR looks fine to me. Lets wait and see what Decode has to say about the multi-monitor behavior. If everything is fine, I'll do another indepth look one last time and merge.

After review the code, I found a serious problem, the slurp one, it maybe 
image

then it will be right, but here use the orgin one, and in slurp, the transform is always normal, I think here should not pass transform into

libwayshot/src/lib.rs
```diff
@@ -93,383 +57,373 @@ pub struct CaptureRegion {
...
+            .map(|output| output.dimensions.y + output.dimensions.height)
+            .max()
+            .unwrap();
+        let capture_region = CaptureRegion {
+            x_coordinate: x1,
+            y_coordinate: y1,
+            width: x2 - x1,
+            height: y2 - y1,
+        };
+        self.screenshot((Transform::Normal, capture_region), cursor_overlay)
```
Decodetalkers, line 422: I think the transform should decided later, by output them self, so this one, Transform I think is needless, when I know a region, I will never know what transform every part is

libwayshot/src/lib.rs
Decodetalkers: this part want to do like this, but from the code, there are some problems, maybe because the original code is little vage

libwayshot/src/lib.rs
c-h-johnson: `new` is very convenient when the consumer does not care about the connection which applies to the majority of usage and I do not see any downsides to making both methods available.

libwayshot/src/lib.rs
Shinyzenith: I see, do you have any suggestions on how to disambiguate this?

c-h-johnson:
> Do we really need to add the anyhow crate?

It prevents using `Result<..., Box<dyn Error>>` in multiple places and greatly improves error handling.

Also all the problems relating to the transform should be fixed now.

Shinyzenith:
Thank you! Given that anyhow improves readability and improves error handling as you mentioned, I am happy to keep it.

@Decodetalkers it seems the fix you mentioned has been merged, would you mind testing it?

Decodetalkers:
> Thank you! Given that anyhow improves readability and improves error handling as you mentioned, I am happy to keep it.
> 
> @Decodetalkers it seems the fix you mentioned has been merged, would you mind testing it?

Yes, this code looks good for me, and transform handle is right, I think, LGTM

Shinyzenith:
> > Thank you! Given that anyhow improves readability and improves error handling as you mentioned, I am happy to keep it.
> > @Decodetalkers it seems the fix you mentioned has been merged, would you mind testing it?
> 
> Yes, this code looks good for me, and transform handle is right, I think, LGTM

What about `screenshot_all` ? Does it composite all screens into one?

Decodetalkers:
> > > Thank you! Given that anyhow improves readability and improves error handling as you mentioned, I am happy to keep it.
> > > @Decodetalkers it seems the fix you mentioned has been merged, would you mind testing it?
> > 
> > 
> > Yes, this code looks good for me, and transform handle is right, I think, LGTM
> 
> What about `screenshot_all` ? Does it composite all screens into one?

Yes, it all works well. Only thing I care is that, since the error is used in lib, I will prefer thiserror, not anyhow, I think anyhow should be used in bin not lib. The crate `thiserror` will provide categorized error, but `anyhow::error` do not have any classification

Shinyzenith:
How do you feel about this^? @c-h-johnson

c-h-johnson:
I think both options are fine so I switched it to `thiserror`.

libwayshot/src/lib.rs
```diff
@@ -1,84 +1,51 @@
+//! `libwayshot` is a convenient wrapper over the wlroots screenshot protocol
+//! that provides a simple API to take screenshots with.
+//!
+//! To get started, look at [`WayshotConnection`].
+
 mod convert;
+mod dispatch;
+mod error;
```
Decodetalkers, line 8: I think it should be pub mod error

Decodetalkers: Because an error is useful, since we use this error, then the error should be public, then users  can use it in pattern match

libwayshot/src/lib.rs
```diff
@@ -93,383 +60,374 @@ pub struct CaptureRegion {
...
+/// # Example usage
+///
+/// ```
+/// let wayshot_connection = WayshotConnection::new().unwrap();
+/// let image_buffer = wayshot_connection.screenshot_all().unwrap();
+/// ```
+#[derive(Debug)]
+pub struct WayshotConnection {
+    conn: Connection,
+    globals: GlobalList,
```
Shinyzenith, line 79:
```suggestion
    pub conn: Connection,
    pub globals: GlobalList,
```

Shinyzenith: Exposing these as public will be needed if the client wants to do other wayland-related operations as you can never recreate the display if you're already attached to it. This would be needed in situations when you're using `WayshotConnection::new`

Shinyzenith (changes requested):

Shinyzenith:
Apart from the above nitpicks, everything else looks fine. The PR is ready for merge once they're addressed.

Shinyzenith:
I think all issues are resolved now. @Decodetalkers would you mind leaving one last approve?

libwayshot/src/lib.rs
Decodetalkers: Oh, I miss it, sorry

Decodetalkers (approved):

Shinyzenith:
Thank you for the pull request! ❤️

Shinyzenith:
Merged as: https://github.com/waycrate/wayshot/commit/8e480ae4cb520e81cc7966cb5ebed37d3483f824

PR #56, opened 2023-08-15
chore: allow pass row fd to capture

Decodetalkers:
https://github.com/pop-os/xdg-desktop-portal-cosmic/blob/master_jammy/src/wayland.rs#L191

from the code in xdg-desktop-portal-cosmis, here should pass a rawfd to it , so maybe expose a function is better?

### DIFF
--- a/libwayshot/src/error.rs
+++ b/libwayshot/src/error.rs
@@ -21,4 +21,10 @@ pub enum Error {
     Global(#[from] GlobalError),
     #[error("connect error: {0}")]
     Connect(#[from] ConnectError),
+    #[error("framecopy failed")]
+    FramecopyFailed,
+    #[error("No supported buffer format")]
+    NoSupportedBufferFormat,
+    #[error("Cannot find Wayland protocols")]
+    ProtocolNotFound(String),
 }
--- a/libwayshot/src/error.rs
+++ b/libwayshot/src/error.rs
@@ -25,6 +25,6 @@ pub enum Error {
     FramecopyFailed,
     #[error("No supported buffer format")]
     NoSupportedBufferFormat,
-    #[error("Cannot find Wayland protocols")]
+    #[error("Cannot find required wayland protocol")]
     ProtocolNotFound(String),
 }
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -13,7 +13,10 @@ mod screencopy;
 use std::{
     cmp,
     fs::File,
-    os::unix::prelude::FromRawFd,
+    os::{
+        fd::{AsFd, AsRawFd},
+        unix::prelude::FromRawFd,
+    },
     process::exit,
     sync::atomic::{AtomicBool, Ordering},
 };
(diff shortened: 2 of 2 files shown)


libwayshot/src/lib.rs
```diff
@@ -142,6 +150,147 @@ impl WayshotConnection {
...
+        // Instantiating screencopy manager.
+        let screencopy_manager = match self.globals.bind::<ZwlrScreencopyManagerV1, _, _>(
+            &qh,
+            3..=3,
+            (),
+        ) {
+            Ok(x) => x,
+            Err(e) => {
+                log::error!("Failed to create screencopy manager. Does your compositor implement ZwlrScreencopy?");
+                panic!("{:#?}", e);
```
Decodetalkers, line 180: e.. I think here should not panic.. if is used in xdg-desktop-portal

Decodetalkers:
let me test it weather it will work

libwayshot/src/lib.rs
Shinyzenith: That is a good point, we can return an error here?

Decodetalkers:
https://github.com/waycrate/wayshot/assets/60290287/a1b81cc8-aea9-4401-8f0e-78733a95fed6
Finshed, so when pass raw fd, it should not use memfd

Decodetalkers:
I think this need https://github.com/waycrate/wayshot/pull/59 , we need to bind wloutput first, and wloutput dispatch too much time

libwayshot/src/error.rs
```diff
@@ -21,4 +21,10 @@ pub enum Error {
     Global(#[from] GlobalError),
     #[error("connect error: {0}")]
     Connect(#[from] ConnectError),
+    #[error("framecopy failed")]
+    FramecopyFailed,
+    #[error("No supported buffer format")]
+    NoSupportedBufferFormat,
+    #[error("Cannot find Wayland protocols")]
```
Shinyzenith, line 28:
```suggestion
    #[error("Cannot find required wayland protocol")]
```

libwayshot/src/lib.rs
```diff
@@ -159,13 +164,26 @@ impl WayshotConnection {
     }
 
     /// Get a FrameCopy instance with screenshot pixel data for any wl_output object.
-    fn capture_output_frame(
+    /// And write data to fd
+    /// It can be used in screencast
```
Shinyzenith, line 168:
```suggestion
```

libwayshot/src/lib.rs
```diff
@@ -159,13 +164,26 @@ impl WayshotConnection {
     }
 
     /// Get a FrameCopy instance with screenshot pixel data for any wl_output object.
-    fn capture_output_frame(
+    /// And write data to fd
```
Shinyzenith, line 167:
```suggestion
    ///  Data will be written to fd.
```

Shinyzenith (changes requested):

Shinyzenith (approved):

PR #58, opened 2023-08-17
chore: show the bug

Decodetalkers:
This pr just use to show what the bug is


Decodetalkers:
emm.. as you have seen, it not run get_outputs, it wll not enter the blocking_dispatch

Decodetalkers:
solved

Issue #62, opened 2023-09-25
Support instantly taking screenshot before choosing portion with slurp


AndreasBackx:
I often want to take a screenshot from a popup window, whether it be in the browser or elsewhere. Unfortunately these popups close when focus is lost. MacOS takes a screenshot when the screenshot button is pressed and the selection tool then allows you to select a portion of the screen thus allowing you to take screenshot of popups.

Having this functionality in Wayshot would not be trivial as it would require creating an overlay on which slurp can then be used. Though is this a feature that could be considered, it could possibly be enabled via a flag like `--instant`?

Shinyzenith:
> Currently `slurp` is also run before Wayshot so we cannot take a screenshot and show it as a moment capture before running slurp so that needs to be changed as well. 

I don't quite understand what this sentence means.

> We could make `--slurp` accept any command and run it from inside of the CLI so we can take a screenshot first and show it to the user after which we can run slurp to select what thay want and then we'd only need to do a crop. What do you think?

Please correct me if I understood you incorrectly:

- In this approach we would have to take the screenshot of all monitors, composite them into one, encode them and then clip/crop it again? Feels inefficient and might lead to loss of detail if I'm not mistaken.

- Leaving the above point aside, what if we only want to select a portion of one monitor? Well now we are still capturing all outputs because we can expect any form of slurp transforms with any number of available monitors being panned across.

- There's also the elephant in the room that I haven't addressed. This would require a major rewrite including `zwlr_layer_shell` and other generic wayland boilerplate that feels out of scope for the project.

# Proposed solution:

Here's a solution that I had thought of since day 1 of writing wayshot which might be more feasible. We take a full-monitor screenshot of the active monitor and then just pipe the contents into a second program ( I had thought of writing a snappy clone which is better..), here we can do: cropping/blurring/coloring etc etc)

I never got the time to execute this idea but if someone wants to pick up on it, let me know.

AndreasBackx:
> I don't quite understand what this sentence means.

When we pass `--slurp`, it uses the output from `slurp` that has already been executed. Though ideally when you select the region you want to save, you want to see exactly what is about to be saved. In this case that means we want to immediately see the screenshot of when we call wayshot. Because slurp runs before this, there is no way we can do that.

So currently it's "slurp -> region screenshot". We would need "screenshot everything -> slurp -> crop".

[...]

PR #63, opened 2023-09-26
feat: add sleep option

Decodetalkers:
add thread sleep to delay screenshot, make it can shot some dialog

Issue: https://github.com/waycrate/wayshot/issues/62

I think it will be ok


Decodetalkers:
and I want to update the deps also

Shinyzenith:
Superceded by #71 and #72

PR #66, opened 2023-10-07
chore: remove unsafe from rawfd

Decodetalkers:
since wayland-rs use borrowedfd, so here do not need unsafe anymore


Decodetalkers:
This fix bad file description, seems I cannot convert owned fd to raw fd

Shinyzenith (approved):

PR #67, opened 2023-10-08
chore: reduce repeat code

Decodetalkers:


libwayshot/src/lib.rs
```diff
@@ -297,123 +323,15 @@ impl WayshotConnection {
...
-                        buffer.destroy();
-                        shm_pool.destroy();
-                        return Ok(frame_format);
-                    }
-                }
-            }
-
-            event_queue.blocking_dispatch(&mut state)?;
-        }
+        // Do the capture part
```
Shinyzenith, line 333:
```suggestion
```

Shinyzenith (changes requested):

Shinyzenith:
Apart from that one comment everything else looks fine. I requested it's removal because the context around the call makes it obvious so it's not needed.

Shinyzenith (approved):

PR #68, opened 2023-10-12
Add tracing, stderr logging, and parallel image alterations

AndreasBackx:
Apologies for the multiple features in a single diff, let me know if you want me to split the diff. This is what has changed:

1. Use [`tracing`](https://docs.rs/tracing/latest/tracing/index.html) instead of `log` as it's more powerful. It allows you to add [instrumentation](https://docs.rs/tracing/latest/tracing/attr.instrument.html) and has a bunch of crates to make telemetry better. 
2. Use `stderr` for logging. At the same time changed it so when using `--stdout`, it does not conflict with `--debug`.
3. In various locations use threads to do operations in paralllel leading to the below ~33% performance improvement for 3 screens. This primarily affects image stuff.

For this last item I tried to use rayon first but there wasn't much difference with a scoped thread implementation besides that we could potentially use it for some work stealing. Though the wins would be minimal for the bigger dependency imo (<10 milliseconds on 300ms) if not it might be worse due to it being optimised for CPU bottlenecks and not IO bottlenecks (we have mostly IO, but the image stuff is CPU).

<details>
<summary>Example output with tracing</summary>

image

</details>

Before:

```
➜ hyperfine "target/release/wayshot --stdout > /dev/null"
Benchmark 1: target/release/wayshot --stdout > /dev/null
  Time (mean ± σ):     310.7 ms ±  16.2 ms    [User: 144.5 ms, System: 105.3 ms]
  Range (min … max):   283.1 ms … 333.4 ms    10 runs
  ```
  
After:
```
➜ hyperfine "target/release/wayshot --stdout > /dev/null"
Benchmark 1: target/release/wayshot --stdout > /dev/null
  Time (mean ± σ):     204.7 ms ±   9.0 ms    [User: 154.6 ms, System: 100.8 ms]
  Range (min … max):   188.2 ms … 216.5 ms    14 runs
```

Iirc there are some slight improvements as well for the part where screenshots are taken, though I don't have stats for that anymore as I tested this on Monday. I was generally doing some performance analysis everywhere. Below is the before and after for a single screen.

Before:

```
➜ hyperfine "target/release/wayshot --stdout --output DP-1 > /dev/null"
Benchmark 1: target/release/wayshot --stdout --output DP-1 > /dev/null
  Time (mean ± σ):      24.2 ms ±   2.1 ms    [User: 8.5 ms, System: 4.6 ms]
  Range (min … max):    21.7 ms …  32.6 ms    104 runs
```

After:

```
➜ hyperfine "target/release/wayshot --stdout --output DP-1 > /dev/null"
Benchmark 1: target/release/wayshot --stdout --output DP-1 > /dev/null
  Time (mean ± σ):      23.9 ms ±   3.3 ms    [User: 9.1 ms, System: 3.9 ms]
  Range (min … max):    21.9 ms …  46.6 ms    61 runs
```

Finally I did some small changes here and there like using a boolean in the library for `cursor_overlay` instead of `i32` (and changing it at the image layer). 

---

If you're interested in the current bottlenecks, they seem to be:

1. Copying the image from the memory mapped file.
2. Overlaying the images on one another
3. Rotating/resizing images
4. Encoding the image

These are all after taking the screen copy though. The screen copy itself is about ~24ms in total last time I checked for 3 monitors. 🤔 The bottleneck seems to be on the server side, we're waiting for the compositor. This was tested on Hyprland.

image

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -129,7 +129,7 @@ impl WayshotConnection {
         ) {
             Ok(x) => x,
             Err(e) => {
-                log::error!("Failed to create ZxdgOutputManagerV1 version 3. Does your compositor implement ZxdgOutputManagerV1?");
+                tracing::error!("Failed to create ZxdgOutputManagerV1 version 3. Does your compositor implement ZxdgOutputManagerV1?");
                 panic!("{:#?}", e);
             }
         };
@@ -154,10 +154,10 @@ impl WayshotConnection {
         }
 
         if state.outputs.is_empty() {
-            log::error!("Compositor did not advertise any wl_output devices!");
+            tracing::error!("Compositor did not advertise any wl_output devices!");
             exit(1);
         }
-        log::debug!("Outputs detected: {:#?}", state.outputs);
+        tracing::debug!("Outputs detected: {:#?}", state.outputs);
         self.output_infos = state.outputs;
 
         Ok(())
@@ -204,8 +204,8 @@ impl WayshotConnection {
         ) {
             Ok(x) => x,
             Err(e) => {
-                log::error!("Failed to create screencopy manager. Does your compositor implement ZwlrScreencopy?");
-                log::error!("err: {e}");
+                tracing::error!("Failed to create screencopy manager. Does your compositor implement ZwlrScreencopy?");
+                tracing::error!("err: {e}");
                 return Err(Error::ProtocolNotFound(
                     "ZwlrScreencopy Manager not found".to_string(),
                 ));
@@ -234,7 +234,7 @@ impl WayshotConnection {
             event_queue.blocking_dispatch(&mut state)?;
         }
 
-        log::debug!(
+        tracing::debug!(
             "Received compositor frame buffer formats: {:#?}",
             state.formats
         );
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by Shinyzenith** [comment]
Hi! thank you for the patch. I haven't gone through all of it just yet. Just to be sure, rayon was replaced with scoped threads right?

I spotted some rayon artifacts in `use` and `Cargo.lock` during a quick overlook. Is that intentional?

File: libwayshot/src/lib.rs [hunk 1]
```diff
@@ -370,63 +374,68 @@ impl WayshotConnection {
         })
     }
 
-    fn create_frame_copy(
+    pub fn create_frame_copy(
```
Shinyzenith, line 377: Does this function need to be public?

```suggestion
    fn create_frame_copy(
```

Shinyzenith, line 377: Does this function need to be public? Is it possibly some artifact from testing?

```suggestion
    fn create_frame_copy(
```

wayshot/src/wayshot.rs
```diff
@@ -31,26 +31,28 @@ where
...
+        Level::TRACE
+    } else {
+        Level::INFO
+    };
+    tracing_subscriber::fmt()
+        .with_max_level(level)
+        .with_writer(std::io::stderr)
+        .init();
 
+    info!("Starting");
```
Shinyzenith, line 44: This is a little unecessary right?

```suggestion
```

Shinyzenith (changes requested):

Shinyzenith:
Apart from those two tiny nit-picks, everything looks fine and the PR performs really well! Thank you for the PR!

Shinyzenith (approved):

PR #69, opened 2023-10-12
refactor(image_composition): image::imageops::{overlay -> replace}

Shinyzenith:
* Overlay takes into account the underlying pixel and the overlayed pixel to interpolate the final pixel data.

  Eg: This takes effect for transparent regions in PNG images.

* Replace performs the same operation for us but instead it does not perform the pixel interpolation and instead just replaces the pixel.

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -19,7 +19,7 @@ use std::{
     thread,
 };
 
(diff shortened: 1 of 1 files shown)


Shinyzenith:
This is better done by PR #71  and #72

PR #72, opened 2023-10-25
Add freeze functionality

AndreasBackx:
Add freeze functionality

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/waycrate/wayshot/pull/72).
* #78
* #77
* #76
* #75
* __->__ #72
* #71

### DIFF
--- a/libwayshot/src/dispatch.rs
+++ b/libwayshot/src/dispatch.rs
@@ -25,11 +25,13 @@ use crate::{
     screencopy::FrameFormat,
 };
 
+#[derive(Debug)]
 pub struct OutputCaptureState {
     pub outputs: Vec<OutputInfo>,
 }
 
 impl Dispatch<WlRegistry, ()> for OutputCaptureState {
+    #[tracing::instrument(skip(wl_registry, qh), ret, level = "trace")]
     fn event(
         state: &mut Self,
         wl_registry: &WlRegistry,
@@ -76,6 +78,7 @@ impl Dispatch<WlRegistry, ()> for OutputCaptureState {
 }
 
 impl Dispatch<WlOutput, ()> for OutputCaptureState {
+    #[tracing::instrument(skip(wl_output), ret, level = "trace")]
     fn event(
         state: &mut Self,
         wl_output: &WlOutput,
@@ -114,6 +117,7 @@ impl Dispatch<WlOutput, ()> for OutputCaptureState {
 delegate_noop!(OutputCaptureState: ignore ZxdgOutputManagerV1);
 
 impl Dispatch<ZxdgOutputV1, usize> for OutputCaptureState {
+    #[tracing::instrument(ret, level = "trace")]
     fn event(
         state: &mut Self,
         _: &ZxdgOutputV1,
@@ -128,19 +132,20 @@ impl Dispatch<ZxdgOutputV1, usize> for OutputCaptureState {
             zxdg_output_v1::Event::LogicalPosition { x, y } => {
                 output_info.dimensions.x = x;
                 output_info.dimensions.y = y;
-                tracing::debug!("Logical position event fired!");
             }
             zxdg_output_v1::Event::LogicalSize { width, height } => {
                 output_info.dimensions.width = width;
                 output_info.dimensions.height = height;
-                tracing::debug!("Logical size event fired!");
             }
+            zxdg_output_v1::Event::Done => {}
+            zxdg_output_v1::Event::Name { .. } => {}
+            zxdg_output_v1::Event::Description { .. } => {}
             _ => {}
         };
     }
 }
 
-/// State of the frame after attemting to copy it's data to a wl_buffer.
+/// State of the frame after attempting to copy it's data to a wl_buffer.
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub enum FrameState {
     /// Compositor returned a failed event on calling `frame.copy`.
@@ -156,6 +161,7 @@ pub struct CaptureFrameState {
 }
 
 impl Dispatch<ZwlrScreencopyFrameV1, ()> for CaptureFrameState {
+    #[tracing::instrument(skip(frame), ret, level = "trace")]
     fn event(
         frame: &mut Self,
         _: &ZwlrScreencopyFrameV1,
@@ -171,7 +177,6 @@ impl Dispatch<ZwlrScreencopyFrameV1, ()> for CaptureFrameState {
                 height,
                 stride,
             } => {
-                tracing::debug!("Received Buffer event");
                 if let Value(f) = format {
                     frame.formats.push(FrameFormat {
                         format: f,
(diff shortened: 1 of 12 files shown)


Decodetalkers:
Libwayshot also used in recording screen, it is not just used for screenshot. So when recording screen, it should not freeze itself,

libwayshot/src/lib.rs
```diff
@@ -174,18 +167,19 @@ impl WayshotConnection {
         cursor_overlay: i32,
         output: &WlOutput,
         fd: T,
-        capture_region: Option<CaptureRegion>,
-    ) -> Result<FrameFormat> {
+    ) -> Result<(FrameFormat, FrameGuard)> {
```
Decodetalkers, line 170: Region is used in recording screen ,I do not think you should remove it

Decodetalkers:
I think we should leave the exported functions have full features of wlr-screencopy, include user can pass region to capture an area. I think region , at least optional region is still needed to be a param

libwayshot/src/lib.rs
```diff
@@ -494,19 +548,32 @@ impl WayshotConnection {
         })
     }
 
+    /// Take a screenshot from the specified region.
+    pub fn screenshot(
+        &self,
+        capture_region: CaptureRegion,
+        cursor_overlay: bool,
+    ) -> Result<RgbaImage> {
+        self.screenshot_region_capturer(RegionCapturer::Region(capture_region), cursor_overlay)
```
Decodetalkers, line 557: Typo?

libwayshot/src/lib.rs
AndreasBackx: You're right, I'll keep support for requesting a screencopy for a specific region for the purposes of screen recording. I hadn't kept that in mind originally.

AndreasBackx:
> I think we should leave the exported functions have full features of wlr-screencopy, include user can pass region to capture an area. I think region , at least optional region is still needed to be a param

I have so far not kept backwards compatibility in mind. Though my plan was to get it to work and think of a good API. Then see what we can do to keep it backwards compatible _for now_. I think these changes teach us how we can still make a breaking API change later, but for now it's best not to break every revision regardless of whether or not we're still in 0.x and we don't promise backwards compatibility. :)

Thanks for the review, will make necessary changes.

libwayshot/src/lib.rs
AndreasBackx: I don't think so? `RegionCapturer` is a new enum to differentiate what we're capturing:

```
pub enum RegionCapturer {
    Outputs(Vec<OutputInfo>),
    Region(CaptureRegion),
    Freeze(Box<dyn Fn() -> Result<CaptureRegion>>),
}
```

Still need to do some documenting however to make this and the reasoning for it more clear!

Decodetalkers:
You bring a file to commit

libwayshot/src/lib.rs
Decodetalkers: Can you add a example above it ?

Decodetalkers:
I think this pr is good.. but if there is some document is better, just no_run is ok

libwayshot/src/lib.rs
AndreasBackx: Will try to add an example of the freeze functionality. :)

Decodetalkers:
I do not think this pr is good enough... I have tried it.

First, I think freeze should  be an option. 
then, I have already use --slurp, but it still capture the whole screen and show it. I think if it is freeze, it should not need to be passed with slurp,

then the biggest problem is scale. when I use 2.0 scale or 1.5, the image cannot overfill my screen, it is terrible

wayshot/src/clap.rs
```diff
@@ -12,10 +12,10 @@ pub fn set_flags() -> Command {
                 .help("Enable debug mode"),
         )
         .arg(
```
Decodetalkers, line 14: I think here should add on optional arg like freeze, because after freeze, we still need to do region select. so at first, we do not need to pass region to it

Decodetalkers:
And you should try to set one of the screen to scale 1.2 or 2, then you will find the problem

Decodetalkers:
and make freeze to an option , another thing is , take screenshots between screens. if we cannot handle scale property, then it will make problem when take screenshot near the side of two screens

Decodetalkers (changes requested):
Sorry, I think this pr still have some problems. thanks for your efforts

wayshot/src/clap.rs
AndreasBackx: I can make a new PR to add something like `--region <GEOMETRY>` which would replicate the old `--slurp`. It would lead to `RegionCapturer.Region(LogicalRegion)` to be used instead of `RegionCapturer.Freeze(...)` and thus do any selection up front. I'd want to do this in a new PR because any changes I make here now will result in me having to do the work again because #76 moves us to Clap derive. Would you want me to add it there perhaps?

AndreasBackx:
> And you should try to set one of the screen to scale 1.2 or 2, then you will find the problem, scale is the most difficult thing to handle, I think , like flameshot cannot handle it propertly.. This feature is like flameshot, if scale can be propertly handle , it will be a greate feature


I have tried setting the scale to 1.2 which gives an error and something I am looking into. Could you explain the issue you see with a scale of 2 as I'm not immediately seeing an issue there? Could you provide the full command you use and the output you get vs what you expected?

wayshot/src/clap.rs
Decodetalkers: I think it's ok. Then solve the scale problem in this pr I think this pr is finished.

Decodetalkers:
Like this kind

Decodetalkers:
> > And you should try to set one of the screen to scale 1.2 or 2, then you will find the problem, scale is the most difficult thing to handle, I think , like flameshot cannot handle it propertly.. This feature is like flameshot, if scale can be propertly handle , it will be a greate feature
> 
> 
> I have tried setting the scale to 1.2 which gives an error and something I am looking into. Could you explain the issue you see with a scale of 2 as I'm not immediately seeing an issue there? Could you provide the full command you use and the output you get vs what you expected?

I have posted the image.. you should scale the captured to size of the screen

AndreasBackx:
> I have posted the image.. you should scale the captured to size of the screen

Thank you! That makes it very clear. Will have a look at it.

libwayshot/src/lib.rs
```diff
@@ -435,30 +411,160 @@ impl WayshotConnection {
...
+                let layer_surface = layer_shell.get_layer_surface(
+                    &surface,
+                    Some(&output_info.wl_output),
+                    Layer::Top,
+                    "wayshot".to_string(),
+                    &qh,
+                    output_info.wl_output.clone(),
+                );
+
+                layer_surface.set_exclusive_zone(-1);
```
Decodetalkers, line 468: You need the size from zxdg_output, not the size from frame_copy, the picture size by zwlr_copy is not always the size of the screen

libwayshot/src/dispatch.rs
```diff
@@ -232,3 +230,43 @@ impl wayland_client::Dispatch<wl_registry::WlRegistry, GlobalListContents> for W
...
+        _conn: &Connection,
+        _qhandle: &QueueHandle<Self>,
+    ) {
+        match event {
+            zwlr_layer_surface_v1::Event::Configure {
+                serial,
+                width: _,
+                height: _,
+            } => {
+                tracing::debug!("Acking configure");
```
Decodetalkers, line 261: I think buffer attach and commit should be here

libwayshot/src/lib.rs
```diff
@@ -435,30 +411,160 @@ impl WayshotConnection {
...
+                surface.commit();
+
+                debug!("Waiting for layer surface to be configured.");
+                while !state.configured_outputs.contains(&output_info.wl_output) {
+                    event_queue.blocking_dispatch(&mut state)?;
+                }
+
+                surface.set_buffer_transform(output_info.transform);
+                surface.set_buffer_scale(output_info.scale);
+                surface.attach(Some(&frame_guard.buffer), 0, 0);
```
Decodetalkers, line 485: before attach, you need to close the buffer, and resize its size to the size of zwloutput

PR #77, opened 2023-11-09
Remove explicit panics and exits

AndreasBackx:
Remove explicit panics and exits

Let's bubble up errors instead of panicking and also not use `exit(...)` because it does not call destructors which might give issues.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/waycrate/wayshot/pull/77).
* #78
* __->__ #77
* #76
* #75
* #72
* #71


Decodetalkers:
I think it is bad to make too many pr at the same time. We cannot know what should be merged first, Can you only mark the first pr to ready to review and other ones to draft?

Shinyzenith (approved):

Shinyzenith:
Changes made in https://github.com/waycrate/wayshot/pull/77/commits/bc9098e6be4ace02c9ce35749f9625c6a765beb2 look fine to me! Approved!

PR #78, opened 2023-11-09
Make region, sizing, and positioning data structs reusable

AndreasBackx:
Make region, sizing, and positioning data structs reusable

We use different structs/fields to store the same three types of data everywhere:
- Position (x,y)
- Size (width,height)
- Region(position, size)

This makes it so they all reuse the same information.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/waycrate/wayshot/pull/78).
* __->__ #78
* #77
* #76
* #75
* #72
* #71

PR #85, opened 2024-02-10
Fix: support correct saving for scaled displays

AndreasBackx:
In #78 it was discussed that master did not support scaled displays. The way screenshotting works is that screencopies are requested in logical regions (so after scaling, a pixel is not a real world pixel), but they're actually saved as "physical" regions (before scaling, aka the unit is 1 real world pixel). Some issues that I fixed were:
1. We created the final image as a "logical" size. So when someone took a screenshot of a 4K screen that was scaled 200%, the picture in memory would be 4K but we would create a 1080p size output file. A 4K output file would now be created.
2. Following the previous step, we need to convert the "logical" region to a "phyiscal" region, aka scale it as well.

But how and where do we scale? From all of the outputs, we take the maximum scale and make sure to scale all other outputs the same amount. This makes sure that our final image will lose the least amount of information. I'm having a hard time explaining this so I'll give some examples in hopes that it makes sense:

1. 1x 1080p + 1x 4K @ 100% -> neither will be resized.
2. 1x 1080p + 1x 4K @ 200% -> the 1080p one will be scaled 200%.
3. 1x 4K @ 100% + 1x 4K @ 200% -> the 4K @ 100% will be scaled 200%.

This seemed the most logical default scaling to me. We could technically make it configurable, though the way it's setup now is that it assumes it will never scale it down.

Still have to do some testing when attaching multiple monitors, but it _should_ work. ™ I am working on my desk setup so it might take me a while to have more than 1 monitor, I'd appreciate someone else testing it with 2 screens and different scaling.

Note: this PR does not fix the freeze overlaying. I actually removed it even because the changes wouldn't allow for it and it needs to be fixed in another PR.

### DIFF
--- a/libwayshot/src/image_util.rs
+++ b/libwayshot/src/image_util.rs
@@ -1,21 +1,24 @@
 use image::{DynamicImage, GenericImageView};
 use wayland_client::protocol::wl_output::Transform;
 
+use crate::region::Size;
+
+#[tracing::instrument(skip(image))]
 pub(crate) fn rotate_image_buffer(
     image: DynamicImage,
     transform: Transform,
-    width: u32,
-    height: u32,
+    logical_size: Size,
+    max_scale: f64,
 ) -> DynamicImage {
     // TODO Better document whether width and height are before or after the transform.
     // Perhaps this should be part of a cleanup of the FrameCopy struct.
-    let (width, height) = match transform {
+    let (logical_width, logical_height) = match transform {
         Transform::_90 | Transform::_270 | Transform::Flipped90 | Transform::Flipped270 => {
-            (height, width)
+            (logical_size.height, logical_size.width)
         }
-        _ => (width, height),
+        _ => (logical_size.width, logical_size.height),
     };
-    let final_image = match transform {
+    let rotated_image = match transform {
         Transform::_90 => image::imageops::rotate90(&image).into(),
         Transform::_180 => image::imageops::rotate180(&image).into(),
         Transform::_270 => image::imageops::rotate270(&image).into(),
@@ -35,14 +38,22 @@ pub(crate) fn rotate_image_buffer(
         _ => image,
     };
 
-    if final_image.dimensions() == (width, height) {
-        return final_image;
+    let scale = rotated_image.width() as f64 / logical_width as f64;
+    // The amount of scaling left to perform.
+    let scaling_left = max_scale / scale;
+    if scaling_left <= 1.0 {
+        tracing::debug!("No scaling left to do");
+        return rotated_image;
     }
 
+    tracing::debug!("Scaling left to do: {scaling_left}");
+    let new_width = (rotated_image.width() as f64 * scaling_left).round() as u32;
+    let new_height = (rotated_image.height() as f64 * scaling_left).round() as u32;
+    tracing::debug!("Resizing image to {new_width}x{new_height}");
     image::imageops::resize(
-        &final_image,
-        width,
-        height,
+        &rotated_image,
+        new_width,
+        new_height,
         image::imageops::FilterType::Gaussian,
     )
     .into()
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -336,7 +336,7 @@ impl WayshotConnection {
     }
 
     /// Get a FrameCopy instance with screenshot pixel data for any wl_output object.
-    #[tracing::instrument(skip_all, fields(output = format!("{output_info}"), region = capture_region.map(|r| format!("{:}", r))))]
+    #[tracing::instrument(skip_all, fields(output = format!("{output_info}"), region = capture_region.map(|r| format!("{:}", r)).unwrap_or("fullscreen".to_string())))]
     fn capture_frame_copy(
         &self,
         cursor_overlay: bool,
@@ -364,7 +364,7 @@ impl WayshotConnection {
             tracing::error!("You can send a feature request for the above format to the mailing list for wayshot over at https://sr.ht/~shinyzenith/wayshot.");
             return Err(Error::NoSupportedBufferFormat);
         };
-        let rotated_size = match output_info.transform {
+        let rotated_physical_size = match output_info.transform {
             Transform::_90 | Transform::_270 | Transform::Flipped90 | Transform::Flipped270 => {
                 Size {
                     width: frame_format.size.height,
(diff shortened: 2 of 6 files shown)


Shinyzenith:
The upscale policy seems acceptable to me. The code from a preliminary view looks fine to me! I don't have a dual monitor system at the moment.

@Decodetalkers Would you mind testing the PR?

Shinyzenith:
> Note: this PR does not fix the freeze overlaying. I actually removed it even because the changes wouldn't allow for it and it needs to be fixed in another PR.

Pardon me, I seem to be losing track due to the large changes recently, what overlaying issue is being discussed here? Is it the image that decode posted in the earlier pr which was due to us not scaling the overlay buffer to physical size?

AndreasBackx:
> > Note: this PR does not fix the freeze overlaying. I actually removed it even because the changes wouldn't allow for it and it needs to be fixed in another PR.
> 
> Pardon me, I seem to be losing track due to the large changes recently, what overlaying issue is being discussed here? Is it the image that decode posted in the earlier pr which was due to us not scaling the overlay buffer to physical size?

There are still two issues with overlaying:
1. Hyprland rotation problem, will need to look into it further and either fix or report a bug.
2. Overlaying in general has not worked yet on a scaled output because this will require one of the newer wlr protocols that allows you to specify that a surface is fractional scaling aware. On a scaled display, it won't overlay correctly afaik because of this. I haven't looked into it much, though we might not even need the special protocol, I'm unsure. Anyhow, overlaying during a freeze is not scaled properly so does not cover the entire output.

Primarily referring to 2 here.

libwayshot/src/lib.rs
```diff
@@ -492,7 +486,7 @@ impl WayshotConnection {
                 }
 
                 surface.set_buffer_transform(output_info.transform);
-                surface.set_buffer_scale(output_info.scale);
+                // surface.set_buffer_scale(output_info.scale());
```
Decodetalkers, line 489: why comment this line? will this line never use again?

libwayshot/src/image_util.rs
```diff
@@ -35,14 +38,22 @@ pub(crate) fn rotate_image_buffer(
         _ => image,
     };
 
-    if final_image.dimensions() == (width, height) {
-        return final_image;
+    let scale = rotated_image.width() as f64 / logical_width as f64;
```
Decodetalkers, line 41: emm.. no , I do not think every picture should resize. 

I think if the dimesions is the same. it should not resize. the final image to file should be clear enough, so I think resize should done in freeze image, not all images

Decodetalkers:
Emm. you should resize in freeze stage, not in finally picture output stage

Decodetalkers:
Maybe you can add a option, for if it is preview image or not, preview image will always resize

AndreasBackx:
> Emm. you should resize in freeze stage, not in finally picture output stage, so I think you just need resize the picture you get in freeze is enough, and maybe this pr is not needed

We should not need to resize at all for a freeze as we'd lose quality. We already have a pixel perfect screenshot we simply need to overlay properly.

Also, we would not be able to anyhow. Freezing is done using the  same buffer as the screenshot and resizing would mean reading it from there which would make the freeze noticeably lag behind the startup.

Decodetalkers:
no, the fix induced many problem, so I create another pr. the size is still not right

libwayshot/src/lib.rs
AndreasBackx: This accepts only an integer and if we were to round it, it'd still be incorrect. It needs fractional scaling support.

libwayshot/src/image_util.rs
AndreasBackx: It's not doing any resizing if it's not needed, see line 44.

AndreasBackx:
@Decodetalkers can you please explain the issues? You created a PR without any information. When looking at the PR you are doing exactly what I said in my previous comment will cause noticeable lag:

> Freezing is done using the same buffer as the screenshot and resizing would mean reading it from there which would make the freeze noticeably lag behind the startup.

You are allocating reading from the existing buffer which takes a lot of time, resizing it, and then writing it to a new buffer. This will lose quality as well and resizing is not at all needed because we have a pixel perfect version of the output in the existing buffers.

Could you explain the issues or possibly point out where my logic is incorrect?

libwayshot/src/image_util.rs
Shinyzenith: This logic looks correct to me. Resolving.

Shinyzenith:
> no, the fix induced many problem

Hi @Decodetalkers, can you kindly enumerate the problems induced so we can improve the PR? From what I can tell, apart from the buffer scale setter function which only accepts integers and not fractionally scaled surfaces -- other things look fine.

Shinyzenith:
> Also, we would not be able to anyhow. Freezing is done using the same buffer as the screenshot and resizing would mean reading it from there which would make the freeze noticeably lag behind the startup.

And this matters heavily as we ideally want the least amount of time from binary invocation to overlay display -- scaling would make us lag behind due to increased allocations.

Decodetalkers:
> @Decodetalkers can you please explain the issues? You created a PR without any information. When looking at the PR you are doing exactly what I said in my previous comment will cause noticeable lag:
> 
> > Freezing is done using the same buffer as the screenshot and resizing would mean reading it from there which would make the freeze noticeably lag behind the startup.
> 
> You are allocating reading from the existing buffer which takes a lot of time, resizing it, and then writing it to a new buffer. This will lose quality as well and resizing is not at all needed because we have a pixel perfect version of the output in the existing buffers.
> 
> Could you explain the issues or possibly point out where my logic is incorrect?

Emmm. I have test your pr, but the size of picture is not correct, the overlay image become too big when the scale is 1.2

Decodetalkers:
you not not adjust the picture on the overlay , I found it still be the size of picture, it will not fully show over the screen. at the beginning, what I concerned is that. the picture size by the wlr-screencopy won't be the size of the screen , so it need to resize to place on the whole screen

Decodetalkers:
https://github.com/waycrate/wayshot/blob/0a8e69d5b8c7bf95735df2b615261774a5cac775/libwayshot/src/lib.rs#L475-L478

I mean the size will always not  be the size of the screen, so this is my problem

Decodetalkers:
I see, this pr is not about freeze overlay, so we are solving two problems

Decodetalkers:
Emm. but without this pr, when use different scale, what will happened? @AndreasBackx I think it will works fine. only I concerned is that the pciture will not cover the screen, the picture shoted will be wrong

libwayshot/src/lib.rs
```diff
@@ -572,8 +577,8 @@ impl WayshotConnection {
                             image_util::rotate_image_buffer(
                                 image,
                                 frame_copy.transform,
-                                frame_copy.frame_format.size.width,
-                                frame_copy.frame_format.size.height,
+                                frame_copy.logical_region.inner.size,
```
Decodetalkers, line 580: Emm. do it need to resize here? I think use the size of the image will be fine, because before, the resize is finally done in the finally output

Decodetalkers: I mean before the logic is shoting all picture, and finally resize them to the size of the selected region, and convert them together. so before the picture do not need to do anything, finally they will all be the size of the selected area, so I cannot understand now the logic is. because after it will be resized again

Decodetalkers:
oh, I see, the logic you use will make the picture be best resized, greate work

Decodetalkers (approved):

Shinyzenith:
I'd like to merge this PR. Before I do so, can anyone confirm that this works on multiple monitor system? I do not have access to one sadly.

Decodetalkers:
it is ok, I will merge it

PR #88, opened 2024-02-16
fix: scale problem

Decodetalkers:


Decodetalkers:
frameguard seems not used

Decodetalkers:
but create a frame will cost lot of time

Decodetalkers:
and people always pass (slurp -d) to select region, so I do not think slurp need args

Decodetalkers:
sorry, seems not fixed

Decodetalkers:
Emm. this still need some adjust, wait

Decodetalkers:
Ok, seems fine

AndreasBackx:
The problem should be fixed by properly scaling the buffer via a Wayland protocol if possible. This will cause the freeze to be slow. @Shinyzenith was looking into this.

Decodetalkers:
do not need anymore

Issue #48, opened 2023-07-03
[feature] clipboard integration


Shinyzenith:
This can be achieved with https://github.com/bugaevc/wl-clipboard

uncomfyhalomacro:
Given that it's a library, I guess this would be easy for you then :eyes:

Shinyzenith:
@uncomfyhalomacro you can try it too

uncomfyhalomacro:
Should clipboard support be an optional feature?

also should it be part of libwayshot or the wayshot cli? not sure where to put it but i feel like wayshot should be the one getting the changes because
- libwayshot is just for screenshot.
- wayshot is the mixture.

Kommynct:
I don't understand the purpose of this feature request at all. It seems like it would just complicate things and add dependencies

wayshot --stdout | wl-copy

works perfectly. Why bother?

Gigas002:
Closing, since this was implemented and we now also have feature flag to disable this behavior.

PR #89, opened 2024-02-23
Clipboard integration

CheerfulPianissimo:
This PR attempts to implement native clipboard support for wayshot using wl-clipboard-rs as discussed in https://github.com/waycrate/wayshot/issues/48

First, I decided to refactor in a new Enum type `SaveLocation` for representing the where the screenshot goes now that it's no longer a boolean choice. I've used `-p` or `--clipboard` as the command line switch for the feature.

wl-clipboard-rs has been added as a dependency and the copying functionality seems to be working with a few caveats I wanted to discuss before proceeding further:
* wl-clipboard-rs only works with compositors using the wlr-data-control protocol, it doesn't work on GNOME's mutter for instance: https://github.com/YaLTeR/wl-clipboard-rs/issues/39
* Reading up on the protocol, it appears that a process can only offer data to be copied only for as long as it stays alive. This is mostly fine for GUI apps but problems arise when you want to copy something to the clipboard and exit as a terminal app. Basically with this PR wayshot will offer up the screenshot for a small moment and immediately exit.
- What this means in practice is that the clipboard functionality in this PR is unusable without a clipboard manager setup. If you have a clipboard manager listening, it will copy the screenshot into it's own memory the moment wayshot offers up the screenshot and the user can get their screenshot from there even after wayshot quits.
- The way to circumvent this is to keep the process alive. wl-clipboard-rs's version of wl-copy for instance does some unsafe shenanigans to fork itself, disconnect stdin/stdout and maintains the image in-memory till some other process overwrites the clipboard: https://github.com/YaLTeR/wl-clipboard-rs/blob/10b35fb2699a0ff65888b1220804bb0c44b65e0f/wl-clipboard-rs-tools/src/bin/wl-copy.rs#L160 I believe this is the same thing wl-copy does. We'll also have to do something similar to achieve parity with the wayshot --stdout | wl-copy method.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,6 +75,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "327762f6e5a765692301e5bb513e0d9fef63be86bbc14528052b1cd3e6f03e07"
 
 [[package]]
+name = "bytecount"
+version = "0.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1e5f035d16fc623ae5f74981db80a439803888314e3a555fd6f04acd51a3205"
+
+[[package]]
 name = "bytemuck"
 version = "1.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -163,6 +169,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive-new"
+version = "0.5.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3418329ca0ad70234b9735dc4ceed10af4df60eff9c8e7b06cb5e520d92c3535"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "dialoguer"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Shinyzenith:
> wl-clipboard-rs only works with compositors using the wlr-data-control protocol, it doesn't work on GNOME's mutter for instance:

That is fine as anything apart from wlroots is out of our project scope.

> Reading up on the protocol, it appears that a process can only offer data to be copied only for as long as it stays alive. This is mostly fine for GUI apps but problems arise when you want to copy something to the clipboard and exit as a terminal app. Basically with this PR wayshot will offer up the screenshot for a small moment and immediately exit.

This is a very well known limitation of the protocol and partly the reason why it was left out of the feature set for a while.

> The way to circumvent this is to keep the process alive. wl-clipboard-rs's version of wl-copy for instance does some unsafe shenanigans to fork itself, disconnect stdin/stdout and maintains the image in-memory till some other process overwrites the clipboard: https://github.com/YaLTeR/wl-clipboard-rs/blob/10b35fb2699a0ff65888b1220804bb0c44b65e0f/wl-clipboard-rs-tools/src/bin/wl-copy.rs#L160 I believe this is the same thing wl-copy does. We'll also have to do something similar to achieve parity with the wayshot --stdout | wl-copy method.

Maybe we can explore this approach if it can be robust enough. I say this from a preliminary overview and naturally I'd need some more code review before I can confirm this.

Shinyzenith:
Before you write any further code. I'd like to kindly request you to rebase your work on freeze-feat-andreas branch as we're currently working on some breaking changes for libwayshot and wayshot.

CheerfulPianissimo:
I think It'll probably be easier to close this PR and do another one on that branch with how changed it is, I'll admit I don't know enough about git convention to be sure of this tho so advice will be appreciated.

I'm also considering having the clipboard copy feature *not* be exclusive of the stdout or file save modes as it is in this PR. That would also make it a feature not trivially synonymous to ` wayshot - | wl-copy`. wayshot can copy to the clipboard and simultaneously save to a file/stdout.

Shinyzenith:
@CheerfulPianissimo Apologies for the late reply, if you're not comfortable with rebase then feel free to open a new pr with the target branch set to the devel branch.

PR #92, opened 2024-02-28
feat: added --human-time flag

rachancheet:
--human-time flag uses more convenient 24-hour format rather than UNIX_EPOCH seconds when saving image to disk.


rachancheet:
if you want date as well , Please let me know .

Shinyzenith:
Thanks for the pr but I am mostly hesitant about adding new crates to the dependency graph. Is there some other options we can explore / features to disable other parts of introduced crate ( I don't have experience with the crate you are using and hence no knowledge of it either )

I like the feature however and the flag name could probably be better named for being intuitive. Also this needs to be documented if we wish to merge this change.

Shinyzenith:
Can you rebase this on feature freeze andreas branch?

PR #93, opened 2024-02-29
feat: added time_stamp flag

rachancheet:
currently default filename is "$UNIX_TIMESTAMP-wayshot.$EXTENSION" . This pr adds flag --time_stamp to use human read-able timestamp instead of seconds from UNIX_EPOCH .

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -43,4 +43,8 @@ pub struct Cli {
     /// Present a fuzzy selector for output/display selection
     #[arg(long, alias = "chooseoutput", conflicts_with_all = ["slurp", "output"])]
     pub choose_output: bool,
+
+    /// Uses time stamp(%H:%M:%S) as file name
+    #[arg(short, long)]
+    pub time_stamp: bool,
 }
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -46,5 +46,5 @@ pub struct Cli {
 
     /// Uses time stamp(%H:%M:%S) as file name
     #[arg(short, long)]
-    pub time_stamp: bool,
+    pub timestamp: bool,
 }
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -43,8 +43,4 @@ pub struct Cli {
     /// Present a fuzzy selector for output/display selection
     #[arg(long, alias = "chooseoutput", conflicts_with_all = ["slurp", "output"])]
     pub choose_output: bool,
-
-    /// Uses time stamp(%H:%M:%S) as file name
-    #[arg(short, long)]
-    pub timestamp: bool,
 }
--- a/wayshot/src/utils.rs
+++ b/wayshot/src/utils.rs
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
 
     format!("{time}-wayshot.{extension}").into()
 }
+
+fn get_hour_minute_from_unix_seconds(seconds: u64) -> String {
+    let total_minutes = seconds / 60;
+
+    let mut current_hour = (((total_minutes / 60) % 24) + 5) % 24;
+
+    let mut current_minute = (total_minutes % 60) + 30;
+    // println!("{}", current_minute);
+    if current_minute > 60 {
+        current_hour += 1;
+    }
+    current_minute = current_minute % 60;
+    if current_hour == 24 {
+        current_hour = 0;
+    }
+
+    // println!("{}", total_minutes as f64 / 60.0);
+
+    format!("{}:{}:{}", current_hour, current_minute, seconds % 60)
+}
+pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
+    let time = match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(n) => get_hour_minute_from_unix_seconds(n.as_secs()),
+        Err(_) => {
+            tracing::error!("SystemTime before UNIX EPOCH!");
+            String::from("")
+        }
+    };
+
+    format!("{time}-wayshot.{extension}").into()
+}
--- a/wayshot/src/utils.rs
+++ b/wayshot/src/utils.rs
@@ -139,28 +139,27 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
         .map(|time| time.as_secs().to_string())
         .unwrap_or("unknown".into());
 
-    format!("{time}-wayshot.{extension}").into()
+    format!("wayshot-{time}.{extension}").into()
 }
 
 fn get_hour_minute_from_unix_seconds(seconds: u64) -> String {
     let total_minutes = seconds / 60;
-
     let mut current_hour = (((total_minutes / 60) % 24) + 5) % 24;
-
     let mut current_minute = (total_minutes % 60) + 30;
-    // println!("{}", current_minute);
+
     if current_minute > 60 {
         current_hour += 1;
     }
+
     current_minute = current_minute % 60;
+
     if current_hour == 24 {
         current_hour = 0;
     }
 
-    // println!("{}", total_minutes as f64 / 60.0);
-
     format!("{}:{}:{}", current_hour, current_minute, seconds % 60)
 }
+
 pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
     let time = match SystemTime::now().duration_since(UNIX_EPOCH) {
         Ok(n) => get_hour_minute_from_unix_seconds(n.as_secs()),
@@ -170,5 +169,5 @@ pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
         }
     };
 
-    format!("{time}-wayshot.{extension}").into()
+    format!("wayshot-{time}.{extension}").into()
 }
--- a/wayshot/src/utils.rs
+++ b/wayshot/src/utils.rs
@@ -1,12 +1,7 @@
 use clap::ValueEnum;
 use eyre::{bail, ContextCompat, Error, Result};
 
-use std::{
-    fmt::Display,
-    path::PathBuf,
-    str::FromStr,
-    time::{SystemTime, UNIX_EPOCH},
-};
+use std::{fmt::Display, path::PathBuf, str::FromStr};
 
 use chrono::{DateTime, Local};
 use libwayshot::region::{LogicalRegion, Position, Region, Size};
@@ -135,15 +130,6 @@ impl FromStr for EncodingFormat {
 }
 
 pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
-    let time = SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .map(|time| time.as_secs().to_string())
-        .unwrap_or("unknown".into());
-
-    format!("wayshot-{time}.{extension}").into()
-}
-
-pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
     let current_datetime: DateTime<Local> = Local::now();
     let formated_time = format!("{}", current_datetime.format("%Y_%m_%d-%H_%M_%S"));
 
--- a/wayshot/src/utils.rs
+++ b/wayshot/src/utils.rs
@@ -8,6 +8,7 @@ use std::{
     time::{SystemTime, UNIX_EPOCH},
 };
 
+use chrono::{DateTime, Local};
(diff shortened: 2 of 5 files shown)


wayshot/src/utils.rs
```diff
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
     format!("{time}-wayshot.{extension}").into()
 }
+
+fn get_hour_minute_from_unix_seconds(seconds: u64) -> String {
+    let total_minutes = seconds / 60;
+
+    let mut current_hour = (((total_minutes / 60) % 24) + 5) % 24;
+
+    let mut current_minute = (total_minutes % 60) + 30;
+    // println!("{}", current_minute);
```
Shinyzenith, line 151: Can we remove these stray comments?

wayshot/src/utils.rs
```diff
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
 
     format!("{time}-wayshot.{extension}").into()
 }
+
+fn get_hour_minute_from_unix_seconds(seconds: u64) -> String {
+    let total_minutes = seconds / 60;
+
+    let mut current_hour = (((total_minutes / 60) % 24) + 5) % 24;
+
```
Shinyzenith, line 149: Nit: remove all stray spaces.

wayshot/src/utils.rs
```diff
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
+    // println!("{}", current_minute);
+    if current_minute > 60 {
+        current_hour += 1;
+    }
+    current_minute = current_minute % 60;
+    if current_hour == 24 {
+        current_hour = 0;
+    }
+
+    // println!("{}", total_minutes as f64 / 60.0);
```
Shinyzenith, line 160: Nit: stray comment

wayshot/src/utils.rs
```diff
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
+
+    format!("{}:{}:{}", current_hour, current_minute, seconds % 60)
+}
```
Shinyzenith, line 163: Spae between functions

wayshot/src/utils.rs
```diff
@@ -141,3 +141,34 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
+    let total_minutes = seconds / 60;
+
+    let mut current_hour = (((total_minutes / 60) % 24) + 5) % 24;
+
+    let mut current_minute = (total_minutes % 60) + 30;
+    // println!("{}", current_minute);
+    if current_minute > 60 {
+        current_hour += 1;
+    }
+    current_minute = current_minute % 60;
```
Shinyzenith, line 155: Nit: Space between if and next line and the if following it too.

Shinyzenith (changes requested):

Shinyzenith:
@Decodetalkers I'm wondering if we should make this the default, UNIX_TIME time isn't a great naming convention. What do you think? 

CC: @AndreasBackx Hi apologies for pinging but would like some of your input if possible -- would changing our naming scheme from wayshot-UNIX_TIME.encoding to wayshot-HUMANIZED_TIME.encoding be considered a breaking change?

AndreasBackx:
@Shinyzenith if you mean an ISO8601 timestamp, then yes. If you mean like "1 day ago", then I'd definitely say no.

Also would make the former the default. I'd also make it wayshot-TIME.EXT so they're grouped if sorted by name. It's technically a breaking change, though we're in 0.x so I think this is fine.

I'm away from a PC, though I'd let people override the name how they want and avoid any further name customisations. Though I think we already have this, cannot check rn.

wayshot/src/utils.rs
rachancheet: pardon , can u please elaborate

rachancheet:
yes user can provide output file name . if no name is provided it saves file in $UNIX_TIMESTAMP-wayshot.$EXTENSION format which is a bit confusing hence this flag.

Shinyzenith:
> @Shinyzenith if you mean an ISO8601 timestamp, then yes. If you mean like "1 day ago", then I'd definitely say no.

Yes I am indeed referring to ISO8601 although this PR isn't formatted with respect to it. I will request that change if everyone agrees on this feature req.
 
> Also would make the former the default. I'd also make it wayshot-TIME.EXT so they're grouped if sorted by name. It's technically a breaking change, though we're in 0.x so I think this is fine.

Please correct me if I am mistaken but at the moment we already do `wayshot-TIME.EXT`.

PS: I don't think changing the naming conventions after we hit 1.x would be acceptable so this will only be entertained for the 0.x period but I will still try to keep it to a low frequency.
We won't hit 1.x until the screencopy protocol either gets accepted in ext namespace of wayland-protocols or zwlr is stabilized to wlr. I think we are fine for now

--- 

@Decodetalkers Can I get an ack for you to make wayshot-ISO8601_ENCODED_TIME.EXT the new default?

rachancheet:
image

wayshot/src/utils.rs
Shinyzenith:
```rs

fn foo() {
}
// enter space here
fn bar() {
}
```

rachancheet:
@Shinyzenith made the requested changes please review.

wayshot/src/cli.rs
```diff
@@ -43,4 +43,8 @@ pub struct Cli {
     /// Present a fuzzy selector for output/display selection
     #[arg(long, alias = "chooseoutput", conflicts_with_all = ["slurp", "output"])]
     pub choose_output: bool,
+
+    /// Uses time stamp(%H:%M:%S) as file name
+    #[arg(short, long)]
+    pub time_stamp: bool,
```
AndreasBackx, line 49:
```suggestion
    pub timestamp: bool,
```

Timestamp is 1 word, `--timestamp` is also better than `--time-stamp`.

wayshot/src/utils.rs
```diff
@@ -139,5 +139,35 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
+
+    format!("{}:{}:{}", current_hour, current_minute, seconds % 60)
+}
+
+pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
+    let time = match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(n) => get_hour_minute_from_unix_seconds(n.as_secs()),
+        Err(_) => {
+            tracing::error!("SystemTime before UNIX EPOCH!");
+            String::from("")
```
AndreasBackx, line 168:
```suggestion
            String::from("unknown")
```

nitpick

wayshot/src/utils.rs
```diff
@@ -139,5 +139,35 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
+    }
+
+    current_minute = current_minute % 60;
+
+    if current_hour == 24 {
+        current_hour = 0;
+    }
+
+    format!("{}:{}:{}", current_hour, current_minute, seconds % 60)
+}
```
AndreasBackx, line 161: This is only printing time `HH:MM:SS` when we need the date and time. Though this actually made me think this format might be quite annoying to use because of the colons and possibly spaces: `DD-MM-YY HH:MM:SS`. Could we go with `2024-12-31 23:59:59`? @Shinyzenith

Additionally, this function can be removed and we could use the `chrono` package. Considering this is a binary, adding a dependency for time stuff is fine.

See the docs here: https://docs.rs/chrono/latest/chrono/format/index.html

AndreasBackx (changes requested):
Looks good, I think we simply need the date as well as that's important. Then `--timestamp` and use `chrono` to make it all easier. Once that's done, I think this is good to go.

wayshot/src/utils.rs
Shinyzenith: > This is only printing time HH:MM:SS when we need the date and time. Though this actually made me think this format might be quite annoying to use because of the colons and possibly spaces: DD-MM-YY HH:MM:SS. Could we go with 2024-12-31 23:59:59? @Shinyzenith

That is fine by me, doesn't seem to invasive and is pretty easy to read / reason through.

> Additionally, this function can be removed and we could use the chrono package. Considering this is a binary, adding a dependency for time stuff is fine.

I initially had suggested not to increase the dependency graph but if you feel so, we can definitely go forth with `chrono`. Apologies for the extra work 😅 @rachancheet .

rachancheet:
image
np problem does this look fine ?

Shinyzenith:
> image np problem does this look fine ?

CC: @AndreasBackx

AndreasBackx:
Apologies for the late reply! (In the future, feel free to ping me on Discord as well.)

And again further apologies, I noticed that my format `DD-MM-YY HH:MM:SS` was not what I intended. I also posted `2024-12-31 23:59:59` which would be `YYYY-MM-DD HH:MM:SS`. The problem with using `wayshot-DD-MM-YYYY HH:MM:SS.ext` is that when sorted alphabetically, it's not sorted by time (which is very much desired imo). Doing `YYYY-MM-SS` does have this benefit which is why I was advocating for it.

Also, now that I see it, should we use `wayshot-2024_03_08-01_01_43.png` instead? This would have the additional benefit that when using this from the terminal, you don't have to quote it.

So do you peeps agree `wayshot-YYYY_MM_DD-HH_MM_SS.ext` would be ideal?

Shinyzenith:
> wayshot-YYYY_MM_DD-HH_MM_SS.ext

I think I would go ahead with this due solely due to the `_`. Quoting files in terminals can be a huge mess.

rachancheet:
image
@Shinyzenith 
@AndreasBackx 
done

Decodetalkers (approved):

wayshot/src/cli.rs
```diff
@@ -43,4 +43,8 @@ pub struct Cli {
     /// Present a fuzzy selector for output/display selection
     #[arg(long, alias = "chooseoutput", conflicts_with_all = ["slurp", "output"])]
     pub choose_output: bool,
+
+    /// Uses time stamp(%H:%M:%S) as file name
```
AndreasBackx, line 47:
```suggestion
    /// Uses `wayshot-YYYY_MM_DD-HH_MM_SS.ext` as a filename.
```

AndreasBackx (approved):
Looks good, thank you! I would say we should make this the default, but we can land this.

Shinyzenith:
I think default would be best. @rachancheet do you agree? If yes we can do so and land the merge.

rachancheet:
image

agreed. What should i rename the flag ??

AndreasBackx:
@rachancheet it should be removed your new function should be the body of the old default function.

wayshot/src/wayshot.rs
```diff
@@ -60,7 +60,13 @@ fn main() -> Result<()> {
...
             }
         }
-        None => Some(utils::get_default_file_name(requested_encoding)),
+        None => {
+            if cli.timestamp {
+                Some(utils::get_time_stamp_file_name(requested_encoding))
+            } else {
+                Some(utils::get_default_file_name(requested_encoding))
+            }
+        }
```
AndreasBackx, line 69: This should be reverted then.

wayshot/src/utils.rs
```diff
@@ -139,5 +140,12 @@ pub fn get_default_file_name(extension: EncodingFormat) -> PathBuf {
...
 
-    format!("{time}-wayshot.{extension}").into()
+    format!("wayshot-{time}.{extension}").into()
+}
+
+pub fn get_time_stamp_file_name(extension: EncodingFormat) -> PathBuf {
+    let current_datetime: DateTime<Local> = Local::now();
+    let formated_time = format!("{}", current_datetime.format("%Y_%m_%d-%H_%M_%S"));
+
+    format!("wayshot-{formated_time}.{extension}").into()
```
AndreasBackx, line 150: And this moved to `get_default_file_name`.

wayshot/src/cli.rs
```diff
@@ -43,4 +43,8 @@ pub struct Cli {
...
+    #[arg(short, long)]
+    pub timestamp: bool,
```
AndreasBackx, line 49: This deleted.

rachancheet:
Done. @AndreasBackx Thanks for the guidance

Decodetalkers (approved):

AndreasBackx (approved):

Shinyzenith (approved):

Shinyzenith:
Hi, I resolved the merge conflicts.

PR #102, opened 2024-03-12
KISS build docs

### LINKED ISSUE #101: build.rs is misused and invoked on any change to sources
Current `build.rs` is misused and slows down iteration time, because it only builds docs and doesn't need to run on _any_ change to sources.

In my case, for around ~3 seconds while I work on #95 and #100 

Solutions are well known:

- [`rerun-if-changed`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) for `docs/*.scd` files
- or to use environment variable to trigger building man pages:

```rs
// build.rs
fn main() {
    if std::env::var("BUILD_DOCS").is_err() {
        std::process::exit(0);
    }

    // build docs if env var is set
}
```

```sh
BUILD_DOCS=1 cargo build --release
```

murlakatamenka:
Fixes #101

- makes iterations while working on `wayshot` faster, thus better for devs and contributors.
- makes `make` more explicit about what it does
- fixes a minor issue of `docs/*.7.gz` not being cleaned


Shinyzenith:
Hi @murlakatamenka. This needs to be rebased onto `freeze-feat-andreas` branch

Shinyzenith:
Make check earlier used to ensure that scdoc could compile the docs too. We use it in CI.

if that can be implemented here then we can go ahead with a merge.

murlakatamenka:
> Make check earlier used to ensure that scdoc could compile the docs too. We use it in CI.
> 
> if that can be implemented here then we can go ahead with a merge.

Previously the docs were built into gzipped man pages during `cargo build`/`build.rs`:

https://github.com/waycrate/wayshot/blob/8f22e6da9005f7495d38d89f2430a2649c4499ef/.github/workflows/build.yml#L19

And also checked in a separate job:

https://github.com/waycrate/wayshot/blob/8f22e6da9005f7495d38d89f2430a2649c4499ef/.github/workflows/build.yml#L32-L43

---

Now that `build.rs` is gone, only documentation job checks if `scdoc` can process the files, this should be enough for CI. Gzipping the output can be considered infallible.

Ideally there should be a job or workflow that gets triggered only if those *.scd files change, because such changes are rare compared modifying the sources.

---

So far I'm done with the changes, feel free to review.

Shinyzenith:
Things look great. Thanks for the PR!

Shinyzenith:
I just realized something I forgot to ask. The code you've written is POSIX compliant right? I would rather keep away from bash extensions.

Decodetalkers:
sorry, now you should push it to master, I accidentally delete the branch. Now that branch is not used anymore

PR #105, opened 2024-03-23
chore: remove useless roundtrip

shivkr6:
There is no need to do roundtrip two times.


Decodetalkers:
@Shinyzenith seems this roundtrip is useless

Shinyzenith (approved):

Issue #106, opened 2024-03-23
Bug: Clipboard flag ignored if path is qualified to file.


Shinyzenith:
![image](https://github.com/waycrate/wayshot/assets/60808802/1dee8963-b3c5-4bf8-b837-ebb3b669169d)

Shinyzenith:
CC: @CheerfulPianissimo

PR #108, opened 2024-03-23
Fix Issue #106: Clipboard flag ignored if path is qualified to file

CheerfulPianissimo:


Shinyzenith:
Hi, please rebase your changes onto freeze-feat-andreas as there's a merge conflict.

Shinyzenith:
Hi it seems you included a commit which already exists in freeze branch. Why?

wayshot/src/wayshot.rs
```diff
@@ -123,40 +124,42 @@ fn main() -> Result<()> {
         }
     };
 
+    let mut buffer = Cursor::new(Vec::new());
```
Shinyzenith, line 127: Lets only create the buffer if clipboard is specified. Else this leads to an extra allocation which we can skip. Lets just call image.save here and create the buffer in the clipboard conditional block.

Shinyzenith (changes requested):

wayshot/src/wayshot.rs
CheerfulPianissimo: I considered that but the stdout block also uses the buffer and I didn't want it to do the buffer write twice in the case of --clipboard + stdout. Should I create the buffer there too?

Shinyzenith: Maybe a function can help

wayshot/src/wayshot.rs
```diff
@@ -128,41 +129,52 @@ fn main() -> Result<()> {
...
-                        MimeType::Autodetect,
-                    )?;
-                }
-            }
+            clipboard_daemonize(buffer)?;
         }
     }
 
+    if !stdout_print && cli.clipboard {
+        let mut buffer = Cursor::new(Vec::new());
```
Shinyzenith, line 144: Do we need the cli.clipboard conditional in the stdout conditional? Is this block outside of the conditional not enough?

```rs

let image_buf: Option<Cursor<Vec<...> = None;
if Some(path) = file {
    /// save
} else {
    /// to stdout

    image_buf = ...;
}

if cli.clipboard && image_buf = Some(image_buffer) {
    //buf stuff
}
```

Shinyzenith (changes requested):

Shinyzenith (changes requested):

wayshot/src/wayshot.rs
```diff
@@ -135,14 +136,18 @@ fn main() -> Result<()> {
...
-        }
+        image_buf = Some(buffer);
     }
 
-    if !stdout_print && cli.clipboard {
-        let mut buffer = Cursor::new(Vec::new());
-        image_buffer.write_to(&mut buffer, requested_encoding)?;
+    if cli.clipboard {
+        let buffer = match image_buf {
+            Some(buf) => buf,
```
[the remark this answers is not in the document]
CheerfulPianissimo: Done!

Issue #109, opened 2024-03-27
Bug: clipboard option freezes everything when used with slurp


Gigas002:
Built from `freeze-feat-andreas` branch. Just running `wayshot --clipboard` works fine, but when trying to run `wayshot --clipboard -s "slurp"` compositor freezes, until wayshot is killed from tty

Shinyzenith:
It's meant to freeze. That's the point of the freeze branch. You're then supposed to select a crop region

Shinyzenith:
I think I misunderstood your issue. Apologies.

Shinyzenith:
CC: @CheerfulPianissimo

CheerfulPianissimo:
Can reproduce. Sway does not actually freeze but you have to full-screen screen apps before you can interact with them. Likely a bad interaction between freeze-feat's use of callbacks+layer-shell and wl-clipboard-rs's thread use or our own daemonization functionality. Will look into it.

CheerfulPianissimo:
My preliminary theory is that freeze-feat is destroys the overlay shell surface as a side effect of the program terminating rather than removing it manually on completion of selection, so when wayshot persists in the background to offer the image on the clipboard, the layer shell does too. Could this be plausible @AndreasBackx ?

CheerfulPianissimo:
yeah that seems to have been it. Have a patch solving this.

AndreasBackx:
@CheerfulPianissimo your assertion seems very likely. It made me look into why we're doing it this way in the first place and I looked into the unsafe code for the clipboard stuff.

@Shinyzenith, I don't fully understand why we need to fork here. Could you explain why we cnanot piggyback on the wl-clipboard implementation? It it because you did not want this to block and it's blocking for a long time?

Also, from my understanding of daemonisation (I'm using it quite extensively at work as a "hack" to make sure terminals exit quickly and can do some work afterwards), this is not daemonisation. We are forking, but we are not "daemonising" the process. That requires a second fork and a setsid call. Unless the Rust stuff does it differently somewhat, the current behaviour simply has 2 processes but should not exit earlier than without this forking code. Though please correct me if I'm wrong.

CheerfulPianissimo:
The rationale for doing the fork is explained here: https://github.com/waycrate/wayshot/pull/89#issue-2151702439 
>  It it because you did not want this to block and it's blocking for a long time?

[...]

PR #110, opened 2024-03-28
feat(clipboard): fix interaction with freeze and select feature: Issue #109

CheerfulPianissimo:
Modifies the overlay creation function to destroy the shell surfaces after use. Not much familiar with layer shell so input from @AndreasBackx would be appreciated on if this is all that is required. I have moved the slurp callback activation into that function too because it needs to be called before the surface is destroyed and destroying the surface needs context only present within this function.

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -413,7 +413,13 @@ impl WayshotConnection {
         Ok(frame_copies)
     }
 
-    fn overlay_frames(&self, frames: &[(FrameCopy, FrameGuard, OutputInfo)]) -> Result<()> {
+    /// Create a layer shell surface for each output,
+    /// render the screen captures on them and use the callback to select a region from them
+    fn overlay_frames_and_select_region(
+        &self,
+        frames: &[(FrameCopy, FrameGuard, OutputInfo)],
+        callback: Box<dyn Fn() -> Result<LogicalRegion, Error>>,
+    ) -> Result<(LogicalRegion)> {
         let mut state = LayerShellState {
             configured_outputs: HashSet::new(),
         };
@@ -491,7 +497,10 @@ impl WayshotConnection {
                 Ok(())
             })?;
         }
-        Ok(())
+        let region = callback()?;
+        layer_shell.destroy();
+        event_queue.blocking_dispatch(&mut state)?;
+        Ok(region)
     }
 
     /// Take a screenshot from the specified region.
@@ -547,7 +556,7 @@ impl WayshotConnection {
             RegionCapturer::Outputs(outputs) => outputs.as_slice().try_into()?,
             RegionCapturer::Region(region) => region,
             RegionCapturer::Freeze(callback) => {
-                self.overlay_frames(&frames).and_then(|_| callback())?
+                self.overlay_frames_and_select_region(&frames, callback)?
             }
         };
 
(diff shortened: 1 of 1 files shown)


Shinyzenith:
Andreas is fairly busy this week, do not expect a fast response.

libwayshot/src/lib.rs
```diff
@@ -413,7 +413,13 @@ impl WayshotConnection {
...
     }
 
-    fn overlay_frames(&self, frames: &[(FrameCopy, FrameGuard, OutputInfo)]) -> Result<()> {
+    /// Create a layer shell surface for each output,
+    /// render the screen captures on them and use the callback to select a region from them
+    fn overlay_frames_and_select_region(
+        &self,
+        frames: &[(FrameCopy, FrameGuard, OutputInfo)],
+        callback: Box<dyn Fn() -> Result<LogicalRegion, Error>>,
+    ) -> Result<(LogicalRegion)> {
```
AndreasBackx, line 422:
```suggestion
    ) -> Result<LogicalRegion> {
```

@Shinyzenith it seems like we don't have clippy enabled in GitHub actions?

```
   Compiling wayshot v1.3.2-dev (/home/andreas/dev/wayshot/wayshot)
warning: unused import: `Region`
  --> libwayshot/src/lib.rs:55:29
   |
55 |     region::{LogicalRegion, Region, Size},
   |                             ^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unnecessary parentheses around type
   --> libwayshot/src/lib.rs:422:17
    |
422 |     ) -> Result<(LogicalRegion)> {
    |                 ^             ^
    |
    = note: `#[warn(unused_parens)]` on by default
help: remove these parentheses
    |
422 -     ) -> Result<(LogicalRegion)> {
422 +     ) -> Result<LogicalRegion> {
    |

warning: unused import: `GenericImageView`
 --> libwayshot/src/image_util.rs:1:27
  |
1 | use image::{DynamicImage, GenericImageView};
  |                           ^^^^^^^^^^^^^^^^

warning: unused variable: `logical_height`
  --> libwayshot/src/image_util.rs:15:25
   |
15 |     let (logical_width, logical_height) = match transform {
   |                         ^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_logical_height`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: `libwayshot` (lib) generated 4 warnings (run `cargo fix --lib -p libwayshot` to apply 3 suggestions)
warning: unused import: `fs::File`
 --> wayshot/src/wayshot.rs:2:5
  |
2 |     fs::File,
  |     ^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: `wayshot` (bin "wayshot") generated 1 warning (run `cargo fix --bin "wayshot"` to apply 1 suggestion)
    Finished release [optimized] target(s) in 0.92s
     Running `target/release/wayshot --clipboard --slurp ''`
Error: error occurred in freeze callback

Location:
    wayshot/src/wayshot.rs:72:9
```

Shouldn't impact the PR, but wanted to mention.

libwayshot/src/lib.rs
```diff
@@ -491,7 +497,10 @@ impl WayshotConnection {
                 Ok(())
             })?;
         }
-        Ok(())
+        let region = callback()?;
+        layer_shell.destroy();
+        event_queue.blocking_dispatch(&mut state)?;
```
AndreasBackx, line 502: If the callback fails, we'll not destroy the layer shell. Can you make it so regardless of the return value, we'll still destroy the layer shell and try to send it on the event queue?

AndreasBackx (changes requested):
Thanks for the fix! Sorry it took me a while to review, I was sick for the entire week. Looks good to land except for the destroying regardless of whetehr the callback succeeds.

libwayshot/src/lib.rs
Shinyzenith: Will address this.

PR #112, opened 2024-03-28
fix(layer-shell-scaling): frozen buffers must scale to output scaling.

Decodetalkers:
I have read the portocol of fractional, the protocal just accept a scale value to client and let client to adjust the scale. so finally the way to scale is done by client. So the scale should be done by ourself

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -14,6 +14,7 @@ mod screencopy;
 use std::{
     collections::HashSet,
     fs::File,
+    io::Write,
     os::fd::AsFd,
     sync::atomic::{AtomicBool, Ordering},
     thread,
@@ -446,6 +447,7 @@ impl WayshotConnection {
             }
         };
 
+        let shm = self.globals.bind::<WlShm, _, _>(&qh, 1..=1, ())?;
         for (frame_copy, frame_guard, output_info) in frames {
             tracing::span!(
                 tracing::Level::DEBUG,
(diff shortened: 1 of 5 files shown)


Decodetalkers:
and now in the code , the way to use slurp is a wrong way

Decodetalkers:
oh.. clippy is merged.. emm, so it is the problem that I have not find my pr

Shinyzenith:
Do we have to do another buffer allocation for the surface layering? Is it not possible to bilinearly scale up/down the current surface contents?

Decodetalkers:
> Do we have to do another buffer allocation for the surface layering? Is it not possible to bilinearly scale up/down the current surface contents?

the size which is used to  scale the buffer only accept int, so when the scale is `int`, I will use the origin one..

Shinyzenith:
> > Do we have to do another buffer allocation for the surface layering? Is it not possible to bilinearly scale up/down the current surface contents?
> 
> the size which is used to scale the buffer only accept int, so when the scale is `int`, I will use the origin one.. but when it is fractional one.. I cannot find out a solution. the protocal about fractional just throw a size to let client to resize it, do you have better idea?

I sadly don't. As far as I'm aware, screencopy already returns to us a pixel-perfect version which is scaled appropriately due to Andreas's patches, we just need a way to hint to the compositor that we have the scaled version already. I'm not sure if my understanding is correct. I'll take a look at the code in the evening and maybe ask in sway-irc if needed.

libwayshot/src/lib.rs
```diff
@@ -446,13 +447,50 @@ impl WayshotConnection {
...
             tracing::span!(
                 tracing::Level::DEBUG,
                 "overlay_frames::surface",
                 output = format!("{output_info}")
             )
             .in_scope(|| -> Result<()> {
+                let buffer = {
+                    let output_size = output_info.logical_region.inner.size;
+                    if output_size != frame_copy.frame_format.size {
+                        let file = tempfile::tempfile().unwrap();
```
AndreasBackx, line 461:
```suggestion
                        let file = tempfile::tempfile()?;
```

If not the correct error type, wrap it in another type, don't unwrap.

libwayshot/src/lib.rs
```diff
@@ -446,13 +447,50 @@ impl WayshotConnection {
...
+
+                        pool.create_buffer(
+                            0,
+                            output_size.width as i32,
+                            output_size.height as i32,
+                            (output_size.width * 4) as i32,
+                            frame_copy.frame_format.format,
+                            &qh,
+                            (),
+                        )
```
AndreasBackx, line 489: There should be no reason to resize. We have a pixel perfect buffer already that represents every single pixel shown on the screen.

Why do you think you need to do this?

wayshot/src/wayshot.rs
```diff
@@ -66,15 +66,11 @@ fn main() -> Result<()> {
...
-    let image_buffer = if let Some(slurp_region) = cli.slurp {
-        let slurp_region = slurp_region.clone();
+    let image_buffer = if cli.slurp {
         wayshot_conn.screenshot_freeze(
             Box::new(move || {
                 || -> Result<LogicalRegion> {
-                    let slurp_output = Command::new("slurp")
-                        .args(slurp_region.split(' '))
-                        .output()?
-                        .stdout;
```
AndreasBackx, line 77: Why change this behaviour? This is removing the ability to pass arguments to slurp.

AndreasBackx (changes requested):
Like @Shinyzenith says (and see my other review comment), the approach is incorrect. We already have a pixel perfect buffer that needs no changes.

Tried this on my machine and it only results in an even blurrier version of what happened before which was already incorrect.

libwayshot/src/lib.rs
Decodetalkers: when the scale is int, I will not resize it, but when it is fractal, I wll resize it manully

wayshot/src/wayshot.rs
Decodetalkers: slurp only accept '-d' to select screen, but -d cannot be passed into, and it is useless

Decodetalkers: image
all the argument is like `-d` `-w`, how can you pass the argument into it? I have try send it , but it can never be received

Decodetalkers: `cargo run -- --slurp '\-d'`
this command will work

but '-d' will never work

PR #120, opened 2024-05-30
Replace nix with rustix

Gigas002:
As title says, this PR proposes replacement of `nix` crate with [rustix](https://github.com/bytecodealliance/rustix).
Currently, the only package dependent on `nix` is a `wl-clipboard-rs`, and it's moving from `nix` to `rustix` as well in upcoming update to match the `wayland-rs`: https://github.com/YaLTeR/wl-clipboard-rs/commit/e81562a06b4c4879bd14524c86c866654a17e4b2
This change simplifies dependency tree for a bit and the migration is pretty straightforward. If there's any downsides I'm not aware of these.


Shinyzenith (approved):

Shinyzenith:
Thank you for your contribution!

PR #128, opened 2024-10-06
fix: I find out a smart way to fix scale problem

Decodetalkers:


Decodetalkers:
I find that the color on my wm is a little strange. on layershell, since the color is Xrgb8888, maybe we should adjust the data in buffer first then do screenshot?

AndreasBackx:
@Shinyzenith would be better for review since I'm lacking in a bit of time. Though this _seems_ fine from first glance. The colour issue should likely fixed in another diff?

libwayshot/src/lib.rs
```diff
@@ -832,6 +835,12 @@ impl WayshotConnection {
                 ));
             }
         };
+        let viewporter = self.globals.bind::<WpViewporter, _, _>(&qh, 1..=1, ()).ok();
+        if viewporter.is_none() {
+            tracing::info!(
+                "Your wm does not support wp_viewporter, so the scale may come out with some problems"
```
Shinyzenith, line 841:
```suggestion
                "Compositor does not support wp_viewporter, display scaling may be inaccurate."
```

Shinyzenith (approved):

libwayshot/src/lib.rs
Shinyzenith: Apart from just the error message, everything else looks good. We can merge

Decodetalkers:
Ok, I will merge it, ☺️

PR #123, opened 2024-06-20
Update image crate

Gigas002:
This PR updates the `image` crate to version [0.25](https://github.com/image-rs/image/blob/main/CHANGES.md).
This version changes the `webp` encoder from binding library to a pure rust native `webp` encoder and also changes `jpeg`-decoders in favor of more performant `zune-jpeg` package.
BTW, maybe it's a good time to introduce the `dependabot` alerts for updating repo dependencies?

PR #129, opened 2025-02-19
Update dependencies

Gigas002:
Rebased to target current `main` branch, updated other deps too (with exception of `image`, breaking changes were introduced only for `gbm`, see: https://github.com/Smithay/gbm.rs/blob/master/CHANGELOG.md)
Supersedes #123


Gigas002:
I suppose, workflow runner requires `libgbm-dev` installation via `apt`?

Shinyzenith:
We have fixed the workflow!

Shinyzenith (approved):

PR #139, opened 2025-03-11
feat(cli): add optional --delay argument for screenshot countdown

suhasamaresh:
Allows specifying a delay before capturing screenshots. Helpful for hovering menus or tooltips.

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -51,4 +51,8 @@ pub struct Cli {
     /// Present a fuzzy selector for output/display selection
     #[arg(long, alias = "chooseoutput", conflicts_with_all = ["slurp", "output"])]
     pub choose_output: bool,
+
+    ///Number of seconds to wait before taking the screenshot
+    #[arg(long, short, default_value = "0", value_name = "SECONDS")]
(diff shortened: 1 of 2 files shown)


Shinyzenith:
Would be best to send notifyd notifications if possible, used when invoked as hotkey binding etc Overall sounds like a nice idea. Just curious as to how thread::sleep would interact with wayland event loop (doesn't look like it should cause any issues). Also accept the sleeping time in milliseconds.

@Decodetalkers Does this look good to you? Do you want this feature to be added?

Decodetalkers:
> Would be best to send notifyd notifications if possible, used when invoked as hotkey binding etc Overall sounds like a nice idea. Just curious as to how thread::sleep would interact with wayland event loop (doesn't look like it should cause any issues). Also accept the sleeping time in milliseconds.
> 
> @Decodetalkers Does this look good to you? Do you want this feature to be added?

I think it is a very good feature, and I want it for a long time

Gigas002 (commented):
When adding new CLI features, please, also provide documentation and a corresponding config file flag

Gigas002:
Superseded by #318

PR #140, opened 2025-03-13
chore: minor improvements in libwayshot lib.rs

OneRandom1509:
- use `ok_or_else()` function to check if frame format exists in `capture_output_frame_get_state_shm()` function
- fixed "failed to create compositor" tracing error message
- fixed typo in `screenshot_single_output()` function's comment

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -526,17 +526,14 @@ impl WayshotConnection {
                         | wl_shm::Format::Bgr888
                 )
             })
-            .copied();
+            .copied()
+            // Check if frame format exists.
+            .ok_or_else(|| {
+                tracing::error!("No suitable frame format found");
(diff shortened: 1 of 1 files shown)


libwayshot/src/lib.rs
```diff
@@ -816,7 +813,7 @@ impl WayshotConnection {
             Ok(x) => x,
             Err(e) => {
                 tracing::error!(
-                    "Failed to create compositor Does your compositor implement WlCompositor?"
+                    "Failed to create compositor. Does your compositor implement WlCompositor?"
```
Decodetalkers, line 816: I think it maybe better to split the commit with that one above

OneRandom1509: sure, will do that!

Decodetalkers (approved):

Decodetalkers:
LGTM

Issue #11, opened 2022-02-27
Support JPEG XL


roland-5:
If I understand correctly, wayshot by default save files as `png` and we can save as `jpg` extension too. I don't know if this is possible or would be enchantment you wanted in your project, but saving screenshots as  JPEG XL (jxl) files would be awesome.

Very nice project!

Shinyzenith:
Yep this is possible and I'm willing to work on it once I gain enough knowledge about encoding with jxl.

Shinyzenith:
https://github.com/image-rs/image/issues/1765
https://github.com/image-rs/image/issues/1765

Related issues. I do not want to do any custom encoding / decoding right now to keep the codebase simple and easy to understand. If it does get implemented in image-rs then I'd be more than happy to expose libwayshot API calls for JXL.

murlakatamenka:
Yeah, JPEG XL is the real deal, especially with the recent [0.10](https://github.com/libjxl/libjxl/releases/tag/v0.10.0) speedups of reference `libjxl` implementation, see here:

https://cloudinary.com/blog/jpeg-xl-and-the-pareto-front

---

My quick tests show that using is really-really fast (Hyprland, wayshot compiled with `-march=native`, Ryzen 5600):

![image](https://github.com/waycrate/wayshot/assets/7361274/0f73bf36-9e2c-4326-9243-408e9fbf36ec)

---

[`zune-image`](https://lib.rs/crates/zune-image) supports jpeg xl, but that's another story, right?

Shinyzenith:
That doesn't seem very fast from the numbers shown -- again I might be mistaken in my interpretation.
I'm generally against adding new crates to the dependency graph unless unavoidable. Does zune support interop with image crate types?

murlakatamenka:
There are a few `zune-` crates behind feature flags for `image-rs`, so there is at least some reference of interop between those.

While jxl is cool, it shines brighter when encoding complex stuff like real world photos in high resolution, like shown in the article. And in lossy but visually lossless scenarios.

---

If we implement png optimization, then jxl gets less attractive for desktop screenshot case. Still an image format to look at, because it's like jpeg, but better in everything, and with lossless mode. Just needs more time and adoption (hello, Google Chrome).

PR #141, opened 2025-03-14
feat: add avif extension support

Gigas002:
[Avif](https://aomedia.org/blog%20posts/avif-meet-the-next-level-image-file-format/) is an open-source, royalty-free high-compression ratio lossless/lossy image format by **Allicance for Open Media**. It's pretty similar to `JPEG-XL`, but since it's support is still lacking in `image-rs` and doesn't seem to appear soon, while we have demand (#11) for similar format and it's very easy to introduce without breaking stuff.

As a downside, if your PC's hardware doesn't support hardware decoding (I suppose something around RDNA3+/RTX40 series GPUs introduced both hardare encoding and decoding for `avif`) it will be hella slow and not recommended. In other case, I personally think it's the best format for both lossy and lossless images storage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
+name = "aligned-vec"
+version = "0.5.0"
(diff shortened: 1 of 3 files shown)


Decodetalkers (approved):

Decodetalkers:
Seems fine to me

Decodetalkers:
Thank you very much!

Shinyzenith:
Hi @Gigas002 Thank you so much for the PR! Would you mind updating the relevant documentation too?

PR #142, opened 2025-03-14
chore: edition 2024 migration

Gigas002:
Not much to say here, mostly automated `cargo fix --edition` stuff


Decodetalkers:
> Wanted to do cargo fmt and try to fix cargo clippy warning in a separate PR. Can do here, just would be a bit hard to compare diffs

We can view every commit separately, so I think that will be ok

Decodetalkers (approved):

Decodetalkers:
seems good to me, I merge it now

Gigas002:
Thanks! Sorry for the wait

PR #145, opened 2025-03-19
feat: file-name-format

Gigas002:
This PR adds optional `file-name-format` argument for CLI. Part of #104. Includes minor (no functional changes) renamings for readability.

The argument is completely optional and default (current) file name without it will be the same (`wayshot-%Y_%m_%d-%H_%M_%S`).

Except for changing the file name format, this PR adds support for using env vars and tilde in paths via `shellexpand` crate.

Supported cases examples:

- no args provided: `wayshot`
- file with encoding: `wayshot 1.jpg`
- file with different encoding: `wayshot 1.jpg --encoding png`
- `file-name-format` provided: `wayshot --file-name-format %H`
- relative dir (must exist or will be treated as file): `wayshot test/`
- relative file: `wayshot test/1.jpg`
- full path: `wayshot /home/user/wayshot/1.jpg`
- env vars: `wayshot $HOME/wayshot/1.jpg`
- tilde: `wayshot ~/wayshot/1.jpg`

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,0 +1,19 @@
+---
+# yamllint disable rule:line-length
+name: check_typos
+
+on:  # yamllint disable-line rule:truthy
+  push:
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: typos-action
+        uses: crate-ci/typos@v1.30.1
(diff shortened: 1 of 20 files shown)


rachancheet:
I recommend using the `--encoding` flag over the extension in the file path when specifying the output format (e.g., `wayshot 1.jpg --encoding png`).

Everything else looks good to me. What do you think, @Shinyzenith ?

Gigas002:
Ah, I overlooked it, thanks! 
Yes, the `requested_encoding` was superior before this change, so it has to be same here, I agree.

Shinyzenith:
> changes) renamings for readability.
> 
> The

Nice catch `--encoding` should be respected first. If it doesn't exist then we auto accept whatever encoding is part of file extension path.

Shinyzenith (approved):

Shinyzenith:
Thank you for the amazing help!

PR #150, opened 2025-03-21
feat: use generics instead every time pass a box

Decodetalkers:
I think use generics will be better, then we do box inside the function, not let user box the function


Shinyzenith:
This sounds like a good idea and it's a breaking change.

Shinyzenith (approved):

Shinyzenith:
Chore: Bump version as it's a breaking change.

PR #154, opened 2025-03-24
feat: implement config file (2)

### LINKED ISSUE #97: Implement config file

Gigas002:
Supersedes #104, closes #97

Most of the stuff, discussed in #104 applied here. We still first take CLI arguments, then config values and only then some defaults to make decisions. Again, by default `config.toml` does not exist and it's OK to have all options commented out/unset.
Default behavior is mostly unchanged, minor non-breaking behavior changes listed below:

- change: `cli.log_level` now is a `Option<Level>` with fallback value `Level::Info` in config. No changes from user perspective whatsoever
- feat: add `cli.config` parameter. Defaults for fallback are also added
- change: in case `cursor` and `clipboard` are `false` (default behavior) -> read values in `config.toml` (defaults for these values are `false` as well, so current users won't notice the difference)
- change: `stdout`, `clipboard` and `FILE` doesn't conflict with each other anymore. Screenshot file will **not** appear on disk now only if no `FILE` provided via args **and** `file` in `config.toml` is `false` at the same time. In all other cases file is saved, with path following this priority: `FILE` > `config.file.path` > current directory. Not breaking, but two category of users affected. First, users currently using `-` as `FILE` will notice that aside from `stdout`, there's a screenshot file in current directory (if no config edits were made). Second catogory is users, that did not provide `FILE` but provided `--clipboard` argument: these users will see similar behavior as above


Shinyzenith (approved):

Shinyzenith:
Things look good to me. The documented toml file is a plus. We should also have a man page section dedicated to the config file. Either that, or we should install the example config in /etc and mention it in man page.

Shinyzenith:
@Decodetalkers can you take a look too?

Decodetalkers:
> @Decodetalkers can you take a look too?

Also seems good for me, we can merge it

Gigas002:
Hey, @Shinyzenith! I have added the doc file for config (took [alacritty's config man](https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd) for reference), but the doc build fails, could you please look at it? I am not familiar with the format, so I don't really know what's wrong

Shinyzenith:
Sure, I'll take a look

Shinyzenith (approved):

PR #157, opened 2025-03-27
Add CLI options to set default config values

noobed-max:
This PR adds new CLI flags to set default values directly in the config file, removes the need for manual editing of config.toml.

New options added:
`--set-default-path <PATH>`: Sets the default screenshot save path.

`--set-cursor <true|false>`: Sets default cursor visibility in screenshots.

`--set-clipboard <true|false>`: Sets default behavior for copying screenshots to the clipboard.

also should I add other set flags too?

### DIFF
--- a/docs/wayshot.1.scd
+++ b/docs/wayshot.1.scd
@@ -84,7 +84,20 @@ Wayshot - Screenshot tool for compositors implementing zwlr_screencopy_v1 such a
 
 	Example: *wayshot --config config.toml*
 
+*--set-default-path <SET_DEFAULT_PATH>*
+	Sets the default screenshot saving path in the config file.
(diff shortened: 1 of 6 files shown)


Decodetalkers:
can you rebase it again? I can review it now

Gigas002:
Honestly I'm against this feature. I just can't come up with any use case where you'd need it instead of `hx ~/.config/wayshot/config.toml`.

I'd understand this if we didn't have `--clipboard`, `--cursor`, etc CLI options and therefore someone who needed scripting support would need to implement something like this, but we already have all we need, so this looks like unneeded implementation for me. I'd prefer not to bloat CLI options without a good reason.

PR #166, opened 2025-04-07
chore: update Nix flake

axelkar:
FYI: you can test these with
```nix
rm -rf target
cargo vendor
# create .cargo/config.toml as `cargo vendor` tells you to
# This tests the build in a pretty isolated environment
nix develop -i --command cargo b
# The bins will fail without the WAYLAND_DISPLAY environment  variable, run them outside of the isolated environment
./target/debug/wayshot
./target/debug/waymirror-egl
```
Should I make a Nix package for wayshot? I'm guessing no since it's already in Nixpkgs and there's really no good reason to use a newer version

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,17 +1,49 @@
 {
-  description = "Development environment for wayshot";
+  description = "wayshot devel and build";
 
-  inputs = {
-    flake-utils.url = "github:numtide/flake-utils";
-  };
+  # Unstable required until Rust 1.85 (2024 edition) is on stable
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-  outputs =
-    {
-      flake-utils,
-      ...
-    }:
-    flake-utils.lib.eachDefaultSystem (system:
-    {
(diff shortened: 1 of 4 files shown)


Shinyzenith:
Hi @axelkar Sadly I do not use nix so I'll tag someone who does. @lavafroth Does this look correct?

flake.nix
```diff
@@ -1,17 +1,49 @@
...
+            nativeBuildInputs = with pkgs; [
+              cargo
+              rustc
+              pkg-config
+
+              rustfmt
+              clippy
+              rust-analyzer
+
+              scdoc
```
lavafroth, line 35: are we actually using scdoc anywhere? I think we could avoid the dependency.

lavafroth (approved):
Just the scdoc nit, otherwise looks good!

flake.nix
lavafroth: Great work. Thanks a lot again.

Shinyzenith (approved):

Shinyzenith:
Thank you for the PR!

PR #179, opened 2025-05-10
feat: upward transfer the features of image

Decodetalkers:


Decodetalkers:
when I use the old version of libwayshot, I found that I cannot save it to png, because in  libwaylib, by default, it disable all the features, and then I cannot even save a png, only when I downgrade my image to 0.24, make the version the same with that on in libwayshot, then I can enable the feature, so here I want to upward transfer the feature, then I can directly enable it in libwayshot

PR #193, opened 2025-06-25
feat: add JPEG-XL encoding support

### LINKED ISSUE #11: Support JPEG XL

Gigas002:
This PR introduces support for encoding [jpeg-xl](https://jpeg.org/jpegxl/), closes #11.
Since there's [no yet support](https://github.com/image-rs/image/issues/1765) for `jxl` in `image-rs` current version temporary uses [jpegxl-rs](https://github.com/inflation/jpegxl-rs) crate.

Note, that since there's no rust-native `jxl` encoder, this crate adds dependency for `libjxl`. I think we may need to make this `[feature]` flag instead?

Also, there are some options available for controlling `jxl` encoder, such as `lossless`, `quality`, `speed`. We can make it configurable via `config.toml` file, though do we really need to go that far? Right now encoder uses default settings, which creates `visualy lossless` results.

One more thing to note, that this PR only adds ability for creating screenshot **file** as `jxl`, for now, at least. The `clipboard` and `stdout` result will be a simple `jpeg`. I am not sure if we should change this behavior, actually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
     - uses: dtolnay/rust-toolchain@stable
       with:
@@ -38,7 +38,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
     - name: Check formatting
       run: |
@@ -49,7 +49,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
     - name: Check docs
       run: |
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: typos-action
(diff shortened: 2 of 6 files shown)


Shinyzenith (dismissed):

EwoutH:
What's the plan with this PR?

Gigas002:
@EwoutH 
It's usable in its current state, except for clipboard screenshot. 
I'm not sure it should be merged rn with this dirty workaround, but at the same time jxl support in image-rs isn't moving at all.
For now, you can use this pr as patch file and build from master branch, with this patch applied, if you want to use this feature.

roland-5:
https://github.com/image-rs/image/issues/1765 and https://github.com/image-rs/image/issues/1979 are closed. The `image` crate should now support jxl via the jxl-oxide crate, if I understand it corretly (https://github.com/image-rs/image/pull/2372 and https://github.com/tirr-c/jxl-oxide/pull/480)

Gigas002:
@roland-5 thanks, I'll check this out!

Gigas002:
Sorry it took a while for me to return to this PR, but I think it should work in all scenarios now.
Unfortunately, jxl-rs only have decoding support, so it doesn't make much sense to use it in our cases, so we have to depend on native libjxl installed on user's machines for now.

Decodetalkers:
ok, let us try this. @Shinyzenith  what do you think of this pr?

Decodetalkers:
ok, I think it should be ok

PR #195, opened 2025-06-26
fix: not work on niri

Decodetalkers:
after commit the surface, seems we just need to do roundtrip not blocking_dispatch

and use Anchor::all to overlay the screen


uncomfyhalomacro:
Tested on niri. Fixed the issue. Have you tested this on sway or river as well? the bug only happens on niri.

uncomfyhalomacro:
> Tested on niri. Fixed the issue. Have you tested this on sway or river as well? the bug only happens on niri.

Tested on river. works. we can merge this.

Shinyzenith:
Hi has this been tested on multimonitors? apart from that I have no idea what the bug was on niri, can someone comment on the same?

CC: @uncomfyhalomacro @Decodetalkers

Decodetalkers:
> Hi has this been tested on multimonitors? apart from that I have no idea what the bug was on niri, can someone comment on the same?
> 
> CC: @uncomfyhalomacro @Decodetalkers 

It shouldn't influence multimonitors, because this change is on the frozen feature. I do not know why it happened on niri, but that place, there is no need to use blocking_dispatch

uncomfyhalomacro:
Leaving more comment, I wonder what was the blocking dispatch for? Is it to prevent creating another surface?

Shinyzenith (approved):

PR #199, opened 2025-08-01
fix: prevent EmbeddedRegion false positive on bordering viewport

Pestdoktor:
Trying to capture e.g. a (0, 123)1920x456 region when there is a second output at (1920, 0) causes the library to request a zero width area of the second output from the compositor which at least on wlroots causes an error.


Decodetalkers (approved):

Decodetalkers:
Can you do some format to fix this ci? Thanks

Pestdoktor:
Done. Sorry for the wait.

Decodetalkers:
Thank you

PR #207, opened 2025-08-16
Add utility functions to support screencasting in xdg-desktop-portal-luminous

nicolo-mn:
This PR adds two utility functions, designed to be used by [`xdg-desktop-portal-luminous`](https://github.com/waycrate/xdg-desktop-portal-luminous):

- `get_available_frame_formats`:  returns a vector of valid formats for a given `WlOutput`. The original idea was to pick a random `WlOutput` from the connection, performing a 1-pixel screenshot to obtain the valid formats using `wlr-screencopy` protocol, and saving them in a `LazyCell`. I preferred to capture the entire screen from a given `WlOutput` passed as parameter instead, to return a vector of `FrameFormat` which contain additional information such as the size and the stride of the buffer. This makes the function more general purpose, allowing it to be used in multiple contexts. For example, right now only 4-bytes pixel formats are supported; if other pixel formats need to be supported in the future, this information could be useful to properly set PipeWire streams' parameters.
- `capture_output_frame_shm_fd_with_format`: same as `capture_output_frame_shm_fd`, but enforces a format to perform the screen capture. Needed since the PipeWire server communicates a preferred video format among the available ones, that must be enforced when capturing video frames.

Additionally, some refactoring was done to avoid code repetitions and `pub` visibility modifier was removed from `capture_output_frame_get_state_shm`, as it was added in an attempt to tackle the issues that `get_available_frame_formats` solves.

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -567,12 +567,7 @@ impl WayshotConnection {
         Ok((state, event_queue, frame))
     }
 
-    // This API is exposed to provide users with access to window manager (WM)
-    // information. For instance, enabling Vulkan in wlroots alters the display
-    // format. Consequently, using PipeWire to capture streams without knowing
-    // the current format can lead to color distortion. This function attempts
-    // a trial screenshot to determine the screen's properties.
-    pub fn capture_output_frame_get_state_shm(
+    fn capture_output_frame_get_state_shm(
         &self,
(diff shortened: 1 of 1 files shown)


Decodetalkers:
@Shinyzenith I think this change is good

Decodetalkers (approved):

Decodetalkers:
I think it will be ok, merged

Issue #94, opened 2024-03-01
feat: Write unit testing suite for wayshot


rachancheet:
Plan is to develop headless integration testing suite using virtual monitors to test screenshots consistency .

Shinyzenith:
Just to be transparent, this will be floated as one of our gsoc ideas.

rachancheet:
originally was assigned to @parag003 but due to prior arrangement he won't be able to work on this issue . I would like to work on this issue

Gigas002:
@Decodetalkers I suppose this may be closed now that #228 had been merhed? Though personally I also would like to see coverage reports, especially on PR changes. I think [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#gitlab-ci) combined with [codecov action](https://github.com/codecov/codecov-action) would be a sufficent solution.

Decodetalkers:
Ok, I think we may also need a coverage ci, @Gigas002 can you help make one?

PR #228, opened 2025-11-09
Add: Unit testing suite for wayshot

xonx4l:
This PR creates unit testing suite for wayshot as discussed in https://github.com/waycrate/wayshot/issues/94

### DIFF
--- a/libwayshot/src/error.rs
+++ b/libwayshot/src/error.rs
@@ -52,9 +52,9 @@ mod tests {
     use super::*;
     use drm::buffer::UnrecognizedFourcc;
     use wayland_client::{
-        backend::{protocol::ProtocolError, WaylandError, InvalidId, ObjectId},
-        globals::{BindError, GlobalError},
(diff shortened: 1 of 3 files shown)


Decodetalkers:
Seems good! Thanks

Decodetalkers (approved):

xonx4l:
@Decodetalkers is there any other issue that you would  suggest me to work on next?

Decodetalkers:
Emm, can you find out a way to make a windows selector? because now we have the windows select protocol, so I decided to add the feature to waysip, and finally add it to wayshot. I want it to be as light as possible, so maybe we still use the cairo to draw it

xonx4l:
I think I can make that . Btw is you available on any social where we can talk and can update on the implementation?

Decodetalkers:
> I think I can make that . Btw is you available on any social where we can talk and can update on the implementation?

You can enter our discord group

PR #242, opened 2025-12-14
Screencast api tidy up, and other things

Decodetalkers:
I want to split screenshot and screencast
and this will also tidy up the api

Now when we use the screencast, we create a new wl_buffer for a new fd every time, this waste a lot of share memory. And with so many needs, the api became quite a mess. 

So now it is time to tidy up it


Decodetalkers:
seems cannot work.. let me find out why

Decodetalkers:
ok. since it is just add some api. so I think it will be no problem. Maybe we can tidy up the api later

Issue #240, opened 2025-12-12
we are still using wlr-screencopy for dmabuf


Decodetalkers:
We need to make it to adapt the ext-image-copy

PR #245, opened 2025-12-17
Enable ext-image-copy

xonx4l:
This PR enables ext-image-copy-dmabuf as discussed in https://github.com/waycrate/wayshot/issues/240


Decodetalkers:
Ohh, maybe it will work. Thank you very much, I will take a try

Decodetalkers:
have you try the waymirror example? it still does not work.. I cannot find out why

Decodetalkers:
the last thing is that the WlBuffer created finally is not fit the requirement of ExtCopyFrame. but the size should be right..

Decodetalkers:
Thank you! I will also try to add a screencast version api for xdg-desktop-portal

PR #248, opened 2025-12-18
Add dmabuf screen logic for screencast part

Decodetalkers:
This also provides a full version waymirror

### DIFF
--- a/libwayshot/src/screencast.rs
+++ b/libwayshot/src/screencast.rs
@@ -19,6 +19,8 @@ use crate::{
     dispatch::{DMABUFState, FrameState, WayshotState},
 };
 
+/// It is a unit to do screencast. It storages used information for screencast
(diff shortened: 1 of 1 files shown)


Decodetalkers:
@Shinyzenith Look, our now is finished! but about the device, how to make llibwayshot automate find the device?

Shinyzenith:
Wow this is awesome! @Decodetalkers I'm not sure how we'd automate device finding tho, will need to look at some other clients I'm guessing

Decodetalkers:
> Wow this is awesome! @Decodetalkers I'm not sure how we'd automate device finding tho, will need to look at some other clients I'm guessing

www. I have read the logic from the xdg-desktop-protal-cosmos and xdp-wlr, and solved it ! no problem now! but the api still is a little mess. I think we need to tidy up them later.

Issue #231, opened 2025-12-01
Add a --preview flag in wayshot CLI


xonx4l:
Before saving or copying to clipboard, open a simple window showing the captured screenshot. Allow the user to confirm (Enter) or discard (Esc).

We can adapt this code from  libwayshot/examples/waymirror.rs
to run before saving/copying.
or 
Could use a lightweight GUI crate .

Suryansh-Dey:
I think it breaks it's philosophy of staying minimalist. If maintainer want, should I add a `--view` option. If specified, it will still save the image but opens it up in his default image viewer, which generally have delete button inbuilt.  
I think passing on preview thing to a previewer app of user's choice, is in line with unix philosophy?

Anshi321:
Hi @xonx4l,

I’m interested in working on this feature and wanted to clarify the preferred direction before starting. The idea of showing a preview window and allowing the user to confirm (Enter) or discard (Esc) makes a lot of sense.

I noticed the reference to libwayshot/examples/waymirror.rs. Would the expectation be to adapt that example directly for a simple preview flow, or would you prefer a more minimal approach using a lightweight GUI crate just for confirmation?

Also, should --preview block until user input before proceeding with save/copy, or should it be optional depending on the output mode?

Once clarified, I can start with a small, focused implementation. Thanks!

Shinyzenith:
Hi, we aren't flameshot and I'd rather not make it that way.

Anshi321:
Understood — thank you for the clarification. 

I’ll focus on a minimal preview flow for the --preview flag, keeping it lightweight and non-GUI-heavy, with simple confirmation before saving or copying.

Diax170:
You can do this with Swappy. Use `wayshot --silent - | swappy -f -`. Then you can either click the save button or close the window. By default it will save to `~/Desktop` but you can change this in `~/.config/swappy/config`:
```ini
[Default]
save_dir=/some/other/save/directory
```

Diax170:
I also made a confirmation dialog with yad and imagemagick:
```shell
#!/usr/bin/env bash

destination="${1:-screenshot.png}"
previewdim="1280x720"

screenshot="$(mktemp XXXXX.png)"
output="$(swaymsg -t get_outputs | jq -r '.[] | select(.focused).name')"
wayshot -o "$output" --silent -- "$screenshot"

preview="$(mktemp XXXXX.png)"
magick "$screenshot" -resize "$previewdim" "$preview"
[...]

PR #249, opened 2025-12-21
Adding --view to view or delete taken SS

Suryansh-Dey:
A very small PR.  
If anything you feel off, please let me know.

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -91,4 +91,8 @@ pub struct Cli {
     ///     3. `None` -- if the config isn't found, the `Config::default()` will be used
     #[arg(long, verbatim_doc_comment)]
     pub config: Option<PathBuf>,
+
+    /// To view taken screen-shot in your image viewer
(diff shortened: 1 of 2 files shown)


Shinyzenith:
Hi this feature feels unnecessary to me. Also this implementation seems incorrect. How do we delete? Where is it saved? If it's a temporary view shouldn't it go through some in memory buffer to a viewer? Why save to disk?

Suryansh-Dey:
> Hi this feature feels unnecessary to me. Also this implementation seems incorrect. How do we delete? Where is it saved? If it's a temporary view shouldn't it go through some in memory buffer to a viewer? Why save to disk?

Actually I have explain this [here](https://github.com/waycrate/wayshot/issues/231#issuecomment-3677525795)

Shinyzenith:
Yeah we won't go ahead with this. Atleast this particular implementation as it assumes a lot of responsibility on xdg-open's part and feels unnecessary in it's current state.

PR #256, opened 2025-12-27
Fix scaling detection for portrait displays

AndreasBackx:
Fixed calculating the needed resize for portrait displays. The `Size` struct passed to `image_util::rotate_image_buffer` was at some moment changed to include the transformation already but wasn't changed there.

Tested using the following setup:
1. Left: 1.25 scale 4K portrait
2. Middle: 1.25 scale 4K landscape
3. Right: 1.25 scale and 4K portrait that was changed to 1.0 scale to see if it worked.

_JPG used below as PNGs were too big to upload here._

### Before

wayshot-2025_12_27-13_05_04

### After
wayshot-2025_12_27-13_05_52

### Before

wayshot-2025_12_27-13_06_57

### After

wayshot-2025_12_27-13_06_52

### DIFF
--- a/libwayshot/src/image_util.rs
+++ b/libwayshot/src/image_util.rs
@@ -7,17 +7,10 @@ use crate::region::Size;
 pub(crate) fn rotate_image_buffer(
     image: DynamicImage,
     transform: Transform,
+    // Includes transform already.
     logical_size: Size,
     max_scale: f64,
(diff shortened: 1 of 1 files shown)


Shinyzenith (approved):

Shinyzenith:
Thanks for the detailed fix! ❤️👏

PR #268, opened 2026-01-08
feature: png-specific optional config section

Gigas002:
Continuation of what I've done for `jxl` encoder in previous PR.

Also updated docfile (forgot to do it for jxl so added that section as well).

Supersedes: #27


Decodetalkers:
Thank you very much!

PR #270, opened 2026-01-15
tidy up apis, mainly the part of egl and waymirror

Decodetalkers:
And use the egl library forked by us

* use the egl library forked by us. because the origin one is not at good maintenance.
* tidy up the waymirror project.
* tidy up the api, make the params follow a good order 
* add easy to use egl screencast api

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Build
       run: |
         rm Cargo.lock
-        cargo build --release
+        cargo build --workspace --release
     - name: Clippy check
       run:
         cargo clippy -- -D warnings
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = ["wayshot", "libwayshot", "./waymirror-egl"]
+default-members = ["wayshot", "libwayshot"]
 
 [workspace.package]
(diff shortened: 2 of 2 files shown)


Decodetalkers:
The color is wired. and there are unhandled error in that original egl lib.

Since egl will only be used in screencast, so can we delete the screenshot api in lib.rs?

Shinyzenith:
Interesting pr, but don't we have a egl backend already? Am I mistaken?

Decodetalkers:
> Interesting pr, but don't we have a egl backend already? Am I mistaken?

Emm. the name of pr is wrong.. I just tidy up the relative api, and use the project that we forked. And then now I tidied up more things..

so mainly , I tidied up the egl api, and tidy up the waymirror, add a screencast api for libwayshot

Decodetalkers:
Now the only problem is 

```
file descriptor expected, object (32), message keymap(uhu)ress <ESC> to quit.
Io error: Invalid argument (os error 22)
Error: Backend(Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" }))
```

Seems the fd leaks.. I do not know why

Decodetalkers:
Ok, I will take responsible for this pr. I will merge it, @Gigas002 thank you for review it

Gigas002:
@Decodetalkers thanks!

PR #277, opened 2026-01-29
feat: add PNG optimization feature using oxipng

### LINKED ISSUE #95: [FR] Optimize png with oxipng
png by default is a reasonable choice. However, it's surely doesn't offer the best compression as of today. For instance, lossless webp consistently beats png in terms of compression ratio.

I suggest adding a separate flag to optimize png output via awesome `oxipng` crate, something like:

```sh
wayshot --extension png --optimize -f /tmp/test.png
```

It is easy via [`oxipng::optimize_from_memory`](https://docs.rs/oxipng/9.0.0/oxipng/fn.optimize_from_memory.html):

```rs
let opts = oxipng::Options::default(); // = preset 2
// let opts = oxipng::Options::from_preset(<u8 from 0 to 6>);
let optimized_png_buf = oxipng::optimize_from_memory(&png_buf, &opts)?;
```

Here is a little script to evaluate the savings:

```sh
mkdir /tmp/wayshot && cd /tmp/wayshot

wayshot -f wayshot.png

for i in {0..6}; do
  time oxipng --opt "$i" --pretend wayshot.png
done
```

Results for a sample [image](https://github.com/waycrate/wayshot/assets/7361274/10d18399-d372-465c-8d19-0f042b09fa84) (`2560x1080`, 414 KiB) on Ryzen 5600:

<details><summary>oxipng 9.0</summary>
<pre>293709 bytes (30.67% smaller)
oxipng --opt "$i" --pretend wayshot.png  0.14s user 0.00s system 99% cpu 0.143 total
---
279314 bytes (34.07% smaller)
oxipng --opt "$i" --pretend wayshot.png  0.31s user 0.01s system 99% cpu 0.322 total
---
274489 bytes (35.21% smaller)
oxipng --opt "$i" --pretend wayshot.png  0.63s user 0.02s system 129% cpu 0.500 total
---
265868 bytes (37.24% smaller)
oxipng --opt "$i" --pretend wayshot.png  2.23s user 0.02s system 264% cpu 0.852 total
---
265107 bytes (37.42% smaller)
oxipng --opt "$i" --pretend wayshot.png  3.77s user 0.03s system 303% cpu 1.252 total
---
265107 bytes (37.42% smaller)
oxipng --opt "$i" --pretend wayshot.png  8.67s user 0.04s system 580% cpu 1.498 total
---
265107 bytes (37.42% smaller)
oxipng --opt "$i" --pretend wayshot.png  12.80s user 0.05s system 781% cpu 1.643 total
</pre>
</details>

Pretty good imo, and using oxipng from Rust itself will be even faster (no CLI args parsing, no buffer copy). 

Reasonable tradeoff of little extra CPU for decent savings in file size even with fast-ish presets 0-2.

---

Surely I can do the plumbing myself, `wayshot --stdout | <best-image-encoder> - <name.ext>`, but I think wayshot and its users will benefit from implementing this feature.

swar09:
closes #95

ran some benchmarks on my VM **i7 14700K 6 Cores** (1680 x 1050) 

|  | Standard PNG | Optimized (`--optimize`) | JXL |
| :--- | :--- | :--- | :--- |
| **Time** | 0.10s | **0.55s** | 0.69s |
| **Memory** | ~35 MB | ~75 MB | **~254 MB** |
| **Size** | 202 KB | **39 KB** | 56 KB |

### why the flag?
while `oxipng` crushes the file size (202kb -> 39kb), the **~5.5x speed hit** makes it kinda unsuitable for a default. 

honestly tho it's a better alternative to JXL right now. in my testing JXL was actually slower (0.69s), ate a ton of ram (~254mb), and the file size was actually bigger (56kb) than the optimized png.

### DIFF
--- a/wayshot/Cargo.toml
+++ b/wayshot/Cargo.toml
@@ -49,6 +49,7 @@ serde = { version = "1.0.219", features = ["derive"] }
 dirs = "6.0.0"
 libwaysip = "0.6.0"
 notify-rust = "4.11.7"
+oxipng = "9.1"
 
 [[bin]]
 name = "wayshot"
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ target
 *.png
 *.ppm
 *.qoi
+*.jxl
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -325,6 +325,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "block2"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -480,6 +492,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+ "terminal_size",
 ]
 
 [[package]]
@@ -550,6 +563,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
 dependencies = [
  "cfg-if",
+]
+
+[[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
+dependencies = [
+ "crossbeam-utils",
 ]
 
 [[package]]
@@ -797,6 +819,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_filter"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bf3c259d255ca70051b30e2e95b5446cdb8949ac4cd22c0d7fd634d89f568e2"
+dependencies = [
+ "log",
+]
+
+[[package]]
+name = "env_logger"
+version = "0.11.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13c863f0904021b108aa8b2f55046443e6b1ebde8fd4a15c399893aae4fa069f"
+dependencies = [
+ "anstream",
+ "anstyle",
+ "env_filter",
+ "log",
+]
+
+[[package]]
 name = "equator"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -879,6 +922,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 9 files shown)


Decodetalkers:
I think also should update the documents

Decodetalkers:
It seems that you did not use the image..emm

I see

swar09 (commented):
LGTM !  all checks passed and the changes look correct

Decodetalkers:
@Gigas002 Can you also take a review? thanks

Shinyzenith:
I do not see the point of a new pr given the old one wasn't merged. I'm not a fan of increasing dependency graph for this tradeoff.

1. We want to be fast.
2. We want to stay slim.
3. If we introduce flags it should be supported on all formats unless it's absolutely unattainable for all

Shinyzenith:
What's the point of supporting compression on one format that leads to a 5.5x slowdown just to save a few kilobytes

swar09:
I picked this up mainly to address #95. In that discussion, adding oxipng seemed like the expected trade-off to get that massive size reduction (~5x smaller) at the cost of some latency (~0.4s slower).

Would you be open to putting this behind a Cargo feature flag? That way the default binary stays fast and lightweight, but the option exists for those who specifically want it.

swar09:
If the extra dependency is a hard blocker even behind a flag, let me know and we can just close this (and likely #95 as well).

wayshot/Cargo.toml
```diff
@@ -49,6 +49,7 @@ serde = { version = "1.0.219", features = ["derive"] }
 dirs = "6.0.0"
 libwaysip = "0.6.0"
 notify-rust = "4.11.7"
+oxipng = "9.1"
```
Gigas002: Please, prefer adding only required features, maybe that'll help to shrink the dependencies list a bit:

> It is recommended to disable the "binary" feature when including Oxipng as a library. Currently, there is no simple way to just disable one feature in Cargo, it has to be done by disabling default features and specifying the desired ones, for example: oxipng = { version = "10.0", features = ["parallel", "zopfli", "filetime"], default-features = false }

https://github.com/oxipng/oxipng?tab=readme-ov-file#library-usage

Also, why not `10.1`? It seems like `10.0+` versions had some perf improvements

Gigas002 (commented):
@swar09 
Thanks for contributing!

I have a mixed feelings for this PR. The implementation itself is good enough, but I am unsure if we need this or not. 

As @Shinyzenith noted, we'd like to not bloat the dependencies tree if possible. Though I'd like to see how it changes if you only add minimal set of required features, maybe then hiding it behind build flag would be an optimal solution.

> If we introduce flags it should be supported on all formats unless it's absolutely unattainable for all

This is also a good point, but I wouldn't be against it if you'll use config file for that instead. We already have `png` and `jxl` specific configurations there. You'd also need to support corresponding options, like `compression` and `filter`, but this shouldn't be complicated as oxipng seems to [support](https://github.com/oxipng/oxipng/blob/master/MANUAL.txt) these as well.

And one more thing, personally I'd also like to see comparison with different config values (like, add default png into comparison table, but with options similar to default oxipng ones: compression level 2, no filter, and so on, you got the gist). Providing actual screenshot files to compare the difference would also help!

P.S. 
One more thing: I’d prefer to see benchmarks run on a bare-metal system rather than a VM if you're comparing against jxl. `libjxl` relies heavily on SIMD instructions like `AVX-512` and `AVX2`. 
Since VMs often mask these flags or use generic CPU profiles, jxl can perform significantly worse in VM than it would on actual hardware, which could skew your comparison data.

swar09:
got it , sure makes sense regarding the avx flags. I will try bare metal and update the PR.

​thanks for the guidance  going to deep dive into the config options now to keep the dependencies minimal and updated.

Gigas002:
@swar09 
Hi! Are you still working on this feature?
We had major refactoring lately, but it should be now easier for you to introduce it as compile-time optional feature.

Gigas002:
I was thinking about this PR for quite a while and was originally planning to merge it for `1.5.0` release, but more I think of it the more I agree with @Shinyzenith.
Adding plain support for image format is one thing, and using built-in library's (in our case, `image`'s `png`) capabilities to tweak the resulting file via config options aligns with it.
However, _optimizing_ the result with completely different tool is kinda out of scope for a _screenshotter_ tool. If users wants to optimize their files, they should run optimizer independently from screenshotter, it's just a completely other field of work.
I am grateful for the contribution, and the request for this had been raised for more than once, but as I explained above, now I consider this out-of-scope for the project.
If @Decodetalkers and @Shinyzenith can agree with me, I'd like to close the PR.

Decodetalkers:
> I was thinking about this PR for quite a while and was originally planning to merge it for `1.5.0` release, but more I think of it the more I agree with @Shinyzenith.
> Adding plain support for image format is one thing, and using built-in library's (in our case, `image`'s `png`) capabilities to tweak the resulting file via config options aligns with it.
> However, _optimizing_ the result with completely different tool is kinda out of scope for a _screenshotter_ tool. If users wants to optimize their files, they should run optimizer independently from screenshotter, it's just a completely other field of work.
> I am grateful for the contribution, and the request for this had been raised for more than once, but as I explained above, now I consider this out-of-scope for the project.
> If @Decodetalkers and @Shinyzenith can agree with me, I'd like to close the PR.

I agree. Maybe should use an extra tools to optimize the target will be better. And also this new feature is just for PNG. So I think @Gigas002 is right

Issue #279, opened 2026-02-03
Silent screenshots with wayshot


DeltaCream:
Is there a way, if I'm using wayshot, to silently take screenshots?

By this I mean, by default, if you use a notification daemon, wayshot sends a notification which is then received by the notifcation daemon.

However, is there a way to just take screenshots without a notifcation sent?

Other screenshot utilities have a way to silently take screenshots such as with the `--silent` or `-s` flag for [Hyprshot](https://github.com/Gustash/Hyprshot).

Would such a thing also be implemented in wayshot, if it hasn't been implemented already?

Shinyzenith:
@Decodetalkers Can you look into this? I think we can just skip the notify-send part

Nuhiat-Arefin:
I think it can be modified in the configuration file. If a CLI flag is needed, I can give it a try

DeltaCream:
Does this mean that the configuration option is already there, or will it just be implemented?

Decodetalkers:
> I think it can be modified in the configuration file. If a CLI flag is needed, I can give it a try

ok ! that will be good. notification has been added recently, and we already have a configure file,and you can set it to disable in config file. but now the document is not enough. and it will also be ok to add a cli flag!

DeltaCream:
Closing the issue now as the changes have been merged. Thanks a lot to @Decodetalkers, @Shinyzenith and @Nuhiat-Arefin for bringing prompt attention to the issue, and to @Suryansh-Dey for their splendid work bringing in the silent flag for wayshot.

I mostly just use this feature for screenshotting the full screen whilst also copying it on the clipboard. I screenshot a lot, so the notifications sent would also be included in the screenshot, which I didn't want, so it was a much needed addition. The only other alternative was to simply not have a notification daemon, which wasn't a good idea as other apps like Discord would break and/or freeze without it.

Here's how I used it in Hyprland:

```
bind = SUPER, Prt Sc, wayshot --clipboard --silent
```

Thanks again!

PR #280, opened 2026-02-04
Adding --silent to cli

Suryansh-Dey:
Addressing issue https://github.com/waycrate/wayshot/issues/279  
Added a --silent CLI flag to wayshot to allow users to suppress notifications without needing to modify their configuration file.

Changes
wayshot/src/cli.rs
Added the silent boolean flag to the Cli struct, at the end.
```rust
/// Silents notification after screenshot
#[arg(long)]
pub silent: bool,
wayshot/src/wayshot.rs
```
Updated the notification logic to priorities the --silent flag.
wayshot.rs
```rust
let notifications_enabled = !cli.silent && base.notifications.unwrap_or(true);
```

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -95,4 +95,8 @@ pub struct Cli {
     ///     3. `None` -- if the config isn't found, the `Config::default()` will be used
     #[arg(long, verbatim_doc_comment)]
     pub config: Option<PathBuf>,
+
+    /// Silents notification after screenshot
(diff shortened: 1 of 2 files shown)


Suryansh-Dey:
I didn't add -s because I thought `s` is such an common word and would be used for something more important.  
Most of the time people would like it to do it through config hence would be a very rare to use flag.

Decodetalkers (approved):

Decodetalkers:
Thank you!

PR #284, opened 2026-02-04
Cache the compilation to avoid recompilation

Suryansh-Dey:
Avoid recompiling every library from scratch, slowing down workflow.  
Caching the /target folder


Decodetalkers:
Thank you!

PR #283, opened 2026-02-04
Adding open file location option in notification

Suryansh-Dey:
## Feature
Now you can click the notification to open file location. Similar to windows. So you can easily rename, move, delete or view it.

## Reason of kernel fork
I did try tokio::spawn but issue is child processes get killed when parent process exits.
Now I may disown the thread but in many cases it creates different issues due to parent getting killed and was too error prone to handle.

This might look scary at first glance but is a very safe way to do it.  
Ready for you feedback.

Issue #204, opened 2025-08-15
stdout still produces a file


uncomfyhalomacro:
# How to reproduce

Assuming you are in `$HOME` and `swappy` installed:

```bash
wayshot - | swappy -f - -o somefilepath.png
```

This saves a regular file while also passing the data as stdout to `swappy`. See video for behaviour.

https://github.com/user-attachments/assets/4427b5e9-6c17-45fd-860c-6eacc3ae1c69

uncomfyhalomacro:
I believe the culprit is in these lines https://github.com/waycrate/wayshot/blob/28331dcc3886f8b9e6bc09f9951fe59744c16acf/wayshot/src/wayshot.rs#L165-L174

Gigas002:
This was made by design, why change again?
The point was to make this behavior more configurable and less conflicting.

> change: stdout, clipboard and FILE doesn't conflict with each other anymore. Screenshot file will not appear on disk now only if no FILE provided via args and file in config.toml is false at the same time. In all other cases file is saved, with path following this priority: FILE > config.file.path > current directory. Not breaking, but two category of users affected. First, users currently using - as FILE will notice that aside from stdout, there's a screenshot file in current directory (if no config edits were made). Second catogory is users, that did not provide FILE but provided --clipboard argument: these users will see similar behavior as above

See: https://github.com/waycrate/wayshot/pull/154

Decodetalkers:
> This was made by design, why change again?
> The point was to make this behavior more configurable and less conflicting.
> 
> > change: stdout, clipboard and FILE doesn't conflict with each other anymore. Screenshot file will not appear on disk now only if no FILE provided via args and file in config.toml is false at the same time. In all other cases file is saved, with path following this priority: FILE > config.file.path > current directory. Not breaking, but two category of users affected. First, users currently using - as FILE will notice that aside from stdout, there's a screenshot file in current directory (if no config edits were made). Second catogory is users, that did not provide FILE but provided --clipboard argument: these users will see similar behavior as above
> 
> See: https://github.com/waycrate/wayshot/pull/154

I see, can you help revert the change back? Thanks

Gigas002:
Not rn (not at PC atm), but yeah, sure. Mostly need to check if it builds and if clippy's happy about the changes

PR #285, opened 2026-02-05
Defaulting to Screenshots directory

Suryansh-Dey:
Changing default saving directory to be `~/Pictures/Screenshots` because all standard tools does this.  
Also nobody would like to fill a random the current working dir to fill with screenshots, scattering them everywhere.

Hence this is lightweight but good default.  
Creating missing parent directories also.

### DIFF
--- a/docs/wayshot.1.scd
+++ b/docs/wayshot.1.scd
@@ -12,7 +12,7 @@ Wayshot - Screenshot tool for compositors implementing zwlr_screencopy_v1 such a
 
 *output*,
 	Location to send captured screenshot to, it can be of the following types:
-		1. A directory, image outputs will be saved in the default format "wayshot-yyyy-mm-dd-hh-mm-ss.png"
+		1. A directory, image outputs will be saved in the default format "wayshot-yyyy-mm-dd-hh-mm-ss.png" (Parent directories will be created if they do not exist)
 
 		2. A path (Encoding is automatically inferred from the extension).
 
--- a/docs/wayshot.5.scd
+++ b/docs/wayshot.5.scd
@@ -91,7 +91,7 @@ This section documents the *[file]* table of the configuration file
 
 	Refer to _[base]_ category's _file_ for examples
 
-	Default: _"None"_ (current working directory)
+	Default: _"None"_ ($HOME/Pictures/Screenshots, fallbacks to current working directory)
 
 *name_format* = _"<string>"_ | _"None"_
 
--- a/wayshot/src/config.rs
+++ b/wayshot/src/config.rs
@@ -87,10 +87,18 @@ pub struct File {
 impl Default for File {
     fn default() -> Self {
         File {
-            path: Some(env::current_dir().unwrap_or_default()),
+            path: Some(File::get_default_screenshot_dir()),
             name_format: Some("wayshot-%Y_%m_%d-%H_%M_%S".to_string()),
             encoding: Some(EncodingFormat::Png),
         }
+    }
+}
+
+impl File {
+    pub fn get_default_screenshot_dir() -> PathBuf {
+        dirs::picture_dir()
+            .map(|path| path.join("Screenshots"))
+            .unwrap_or_else(|| env::current_dir().unwrap_or_default())
     }
 }
 
--- a/wayshot/src/wayshot.rs
+++ b/wayshot/src/wayshot.rs
@@ -1,6 +1,6 @@
 use config::Config;
 use std::{
-    env,
+    fs,
     io::{self, BufWriter, Cursor, Write},
 };
 
@@ -101,7 +101,7 @@ fn main() -> Result<()> {
             if base.file.unwrap_or_default() {
                 let dir = file
                     .path
-                    .unwrap_or_else(|| env::current_dir().unwrap_or_default());
+                    .unwrap_or_else(config::File::get_default_screenshot_dir);
                 Some(utils::get_full_file_name(&dir, &file_name_format, encoding))
             } else {
                 None
@@ -262,6 +262,9 @@ fn main() -> Result<()> {
             let mut image_buf: Option<Cursor<Vec<u8>>> = None;
 
             if let Some(f) = file {
+                if let Some(parent) = f.parent() {
+                    fs::create_dir_all(parent)?;
+                }
                 if encoding == EncodingFormat::Jxl {
                     if let Err(e) = utils::encode_to_jxl(
                         &image_buffer,


Decodetalkers:
I cannot decide it. But I prefer this pr. I want to hear the opinion from @Shinyzenith . Wait for him when he is free

Gigas002:
> Changing default saving directory to be ~/Pictures/Screenshots because all standard tools does this.

Can you please also list which tools does that, for the sake of information? I always override default saving locations to my custom folder so I'm not aware about other tools's defaults, but KDE's default screenshotting tool, Spectacle, seems to save into `~/Pictures` if I've [googled correctly](https://docs.kde.org/stable_kf6/en/spectacle/spectacle/ch04s02.html).

I prefer this one over #283 (no new `unsafe` stuff introductions better than some, even though we already use `fork` for clipboard), but still personally I dislike that it introduces breaking behavior for current users.. 
By the looks of it, it should work, but can you please confirm, that running something like `wayshot screenshot.png` will create screenshot in **current** directory?

Also some related documentation changes are required as well (e.g. wayshot.5.scd).

Suryansh-Dey:
> Can you please also list which tools does that, for the sake of information

I have used windows and KDE default screenshot tools and both save them in `~/Pictures/Screenshot`  

> I prefer this one over https://github.com/waycrate/wayshot/pull/283 (no new unsafe stuff introductions better than some, even though we already use fork for clipboard)

We are calling notification function or forking it, only at the end of the main function. So it's very unlikely to get into any issues with forking. But still it's two word of unsafe.
> By the looks of it, it should work, but can you please confirm, that running something like wayshot screenshot.png will create screenshot in current directory?

wayshot-2026_02_06-12_32_25

Suryansh-Dey:
Ready for feedback.

Shinyzenith:
Hey not sure if I want this change because it's not backwards compatible. I'd rather not break user scripts.

What do you think @Decodetalkers ?

Decodetalkers:
> Hey not sure if I want this change because it's not backwards compatible. I'd rather not break user scripts.
> 
> What do you think @Decodetalkers ?

I see. Yes. It may break user scripts.. or maybe we can add a cli flag to control it? Like something save to directory?

Decodetalkers:
> Hey not sure if I want this change because it's not backwards compatible. I'd rather not break user scripts.
> 
> What do you think @Decodetalkers ?

I have thought about this pr. Maybe we should make breaking change, and write it in document. The saved picture in current folder is with random name, and is not good for use. But it was designed that even with stdout, a file will still be saved in current directory. https://github.com/waycrate/wayshot/issues/204

but the option can be closed. So this pr will be safe I think

Suryansh-Dey:
I have a nice idea! Introduce a flag, to avoid breaking change, like -d, which will just take a file name, no need to mention the full path.  
The screenshot will be saved in default dir or dir specified by user config. This saves user time when wanting to store image of desired name since he won't have to mention full path anymore.

Also he can mention `-d /private/water.png` to store it in a sub-folder of picture folder.

Decodetalkers:
> I have a nice idea! Introduce a flag, to avoid breaking change, like -d, which will just take a file name, no need to mention the full path. The screenshot will be saved in default dir or dir specified by user config. This saves user time when wanting to store image of desired name since he won't have to mention full path anymore.
> 
> Also he can mention `-d /private/water.png` to store it in a sub-folder of picture folder.

I think it is a good idea! @Shinyzenith @Gigas002 What do you think about it? If you agree with it, what name should we name this flag?

Gigas002:
I like the idea, though this behavior exists already in a bit different form, if I understood you correctly:

```sh
wayshot --file-name-format "screenshot" ~/downloads
```

above command will create file `~/downloads/screenshot.encoding` (with encoding provided via `--encoding` option or using `config`/default value automatically). A bit nasty way to do it, but I guess it is the same as your idea? 

This also works with non-full path:

```sh
# create screenshot.encoding in current dir
wayshot --file-name-format "screenshot" .
# or create in existing subdir
wayshot --file-name-format "screenshot" scr/
```

The only difference, it does not create parent directories if provided path doesn't exist, so:

```sh
wayshot --file-name-format "screenshot" ~/downloads/screenshots/scr
2026-02-14T05:06:16.333251Z ERROR wayshot: Failed to encode to JXL: No such file or directory (os error 2)
```

this fails for me because `screenshots` dir does not exist. So maybe it'd be better to edit this behavior instead?

Suryansh-Dey:
But won't changing an existing flag's behavior be a breaking change? Actually I meant, if dir is not mentioned, it should use default dir from config or `~/Pictures/Screenshots`

Gigas002:
Depends on what changes do you want to implement. Adding parent directories automatic creation sounds more like a bugfix to me rather than breaking api.

> Actually I meant, if dir is not mentioned, it should use default dir from config or ~/Pictures/Screenshots

Yes. already working this way:

```sh
wayshot --file-name-format "screenshot"
```

will create a screenshot file in a default directory, specifies in config. E.g. my config:

```toml
[base]
clipboard = true
file = true

[file]
path = "$HOME/images/screenshots"
name_format = "%Y-%m-%d_%H-%M-%S"
encoding = "jxl"
```

And the above command, without specified dir path, will create file: `~/images/screenshots/screenshot.jxl`

Suryansh-Dey:
Okay so I will make this parent dir creation. But won't it be more convenient to have a flag such that  `-f "water"` saves an image of water.encoding at my configured dir? and `-f /world/water` will create image `default_dir/world/water.encoding`?

Suryansh-Dey:
> Depends on what changes do you want to implement. Adding parent directories automatic creation sounds more like a bugfix to me rather than breaking api.
> 
> > Actually I meant, if dir is not mentioned, it should use default dir from config or ~/Pictures/Screenshots
> 
> Yes. already working this way:
> 
> ```shell
> wayshot --file-name-format "screenshot"
> ```
> 
> will create a screenshot file in a default directory, specifies in config. E.g. my config:
> 
> ```toml
> [base]
> clipboard = true
> file = true
> 
> [file]
> path = "$HOME/images/screenshots"
> name_format = "%Y-%m-%d_%H-%M-%S"
> encoding = "jxl"
> ```
> 
> And the above command, without specified dir path, will create file: `~/images/screenshots/screenshot.jxl`

Wow! Okay means I just need to create parent dir right? I think I have already addressed that in this PR. See the files change and scroll down to bottom.

Gigas002:
> Wow! Okay means I just need to create parent dir right?

Yeah, sounds this way.
I think the best possible fix would be to check the prepared path and create dir on that step, rather than implement this change only for flag:

https://github.com/waycrate/wayshot/blob/6f570566b7f98e2837628383d70b45806827c0f2/wayshot/src/wayshot.rs#L86C11-L109C11

Suryansh-Dey:
> wayshot --file-name-format "screenshot" ~/downloads
> 
> above command will create file ~/downloads/screenshot.encoding (with encoding provided via --encoding option or > using config/default value automatically). A bit nasty way to do it, but I guess it is the same as your idea?

Sorry but I am a bit confused and needed a bit of help.  Shouldn't below command error or create a dir of name `/lol/`?
wayshot-2026_02_16-19_03_56

Gigas002:
Hmm. If dir doesn't exist on provided path, there's no way to tell if it's intended to be a dir or a file, and therefore we always force this as filename:

https://github.com/waycrate/wayshot/blob/8460699196956ce7865dd32cee24fe9e63d8c845/wayshot/src/utils.rs#L169

Ugh, I wonder what would be a proper solution in this situation and what other members think. 
We either change the defaults like this PR proposes, either implement some feature (which really does looks to me as one more source to confuse users -- having 3 different, partially conflicting parameters for path is clearly wrong), either tell users to change the config value to `~/Pictures/Screenshots` by their own.

PR #287, opened 2026-02-06
fix: add missing doc entries

Gigas002:
Noticed that after latest changes some cli/config options were not listed and added them.


Decodetalkers (approved):

Decodetalkers:
Thank you

PR #294, opened 2026-02-16
refactor: replace unwrap() with proper error handling in libwayshot

0xsamalt:
Changes I have made:
- Replaced `try_into().unwrap()` with a match statement in **dispatch.rs** to gracefully handle malformed data from the compositor.
- Converted `Card::open` to return a `std::io::Result<Self>` and updated its call sites (including lib.rs)
- I have refactored the internal format handling to use direct indexing (`state.formats[0]`) where values are guaranteed, reducing the reliance on `unwrap()` while maintaining intended fail-fast behavior for logic bugs.


Decodetalkers (approved):

Decodetalkers:
Thank you!

Issue #301, opened 2026-02-20
Refactor wayshot.rs


Gigas002:
The more features we add/change behavior of wayshot.rs, more painful it becomes to read it.
We should refactor it a bit, separating `main` on several functions would mostly help IMO.

Gigas002:
I think I have some kind of a roadmap for this in my head, going to work on this one. Probably will draft some version tomorrow.

PR #302, opened 2026-02-21
feat: add format choice for color option

0xsamalt:
Okay, I have made the following changes:

- Added a --color-format flag that outputs a single "raw" string to stdout, allowing users to pipe color codes directly into tools like wl-copy without grep or awk.

- Implemented native HSL conversion and multiple hex/RGB variants (hex, hex-alpha, rgb, rgba), all manageable via a new ColorFormat enum.

- Maintained full backward compatibility by defaulting to the original multi-line plain output when the format flag is omitted.

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -6,6 +6,7 @@ use clap::{
         Styles,
         styling::{AnsiColor, Effects},
     },
+    ValueEnum,
 };
 use tracing::Level;
 
@@ -17,6 +18,17 @@ fn get_styles() -> Styles {
         .usage(AnsiColor::Green.on_default() | Effects::BOLD)
         .literal(AnsiColor::Blue.on_default() | Effects::BOLD)
         .placeholder(AnsiColor::Green.on_default())
+}
+
+#[derive(ValueEnum, Clone, Debug, Default, PartialEq)]
+pub enum ColorFormat {
+    #[default]
+    Plain,
+    Hex,
+    HexAlpha,
+    Rgb,
+    Rgba,
+    Hsl,
 }
 
 #[derive(Parser)]
@@ -70,6 +82,12 @@ pub struct Cli {
     /// Grasp a point in screen and get its color
     #[arg(long, conflicts_with_all = ["geometry", "output", "choose_output"])]
     pub color: bool,
+
+    /// Format for color output when using --color.
+    /// Available options: plain, hex, hex-alpha, rgb, rgba, hsl
+    /// If omitted, falls back to plain (prints all formats).
+    #[arg(long, value_name = "FORMAT", requires = "color", default_value = "plain")]
+    pub color_format: ColorFormat,
 
     /// Capture a specific toplevel window by name ("app_id title").
     #[arg(long, alias = "window", conflicts_with_all = ["geometry", "output", "choose_output", "choose_toplevel"])]
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -1,12 +1,11 @@
 use std::path::PathBuf;
 
 use clap::{
-    Parser,
+    Parser, ValueEnum,
     builder::{
         Styles,
         styling::{AnsiColor, Effects},
     },
-    ValueEnum,
 };
 use tracing::Level;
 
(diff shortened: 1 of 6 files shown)


wayshot/src/cli.rs
```diff
@@ -71,6 +82,17 @@ pub struct Cli {
     #[arg(long, conflicts_with_all = ["geometry", "output", "choose_output"])]
     pub color: bool,
 
+    /// Format for color output when using --color.
+    /// Available options: plain, hex, hex-alpha, rgb, rgba, hsl
+    /// If omitted, falls back to plain (prints all formats).
+    #[arg(
```
Decodetalkers: can it be a sub option of color? use parse to parse a subcommand?

0xsamalt: hmm... that would be better

Shinyzenith:
Why do I feel like we are reaching flag hell?

Decodetalkers:
> Why do I feel like we are reaching flag hell?

but this pr now does not induce new flag for the Cli structure

Shinyzenith:
> > Why do I feel like we are reaching flag hell?
> 
> but this pr now does not induce new flag for the Cli structure

it's called --color-format?

Decodetalkers:
> > > Why do I feel like we are reaching flag hell?
> > 
> > 
> > but this pr now does not induce new flag for the Cli structure
> 
> it's called --color-format?

Now it is just add an extra option for --color

for example 
```
wayshot-- --color rgb
```

@0xsamalt by the way, I think you should also update the document. @Gigas002 What do you think about this? I hope you also take a look

Gigas002:
I think the implementation itself is good enough, not much to add here. I'd move `rgb_to_hsl` into `utils.rs`, but that could be done as part of #301 , so it's not a big dealbreaker. But what I am not sure, is if we even need color picker's functionality in wayshot.
I know, it's not uncommon for screenshoting tools to be all-in-one applications, like `flameshot` or `sharex`, etc, but personally I'd like it to stay UNIX'ey slim and focused on one thing: screenshotting, like `grim` does. That's not my decision to make though, just an opinion.
Lately I've been feeling that we're kinda losing the goal to be light and performant screenshotting tool and I'd like to discuss it with other team members. We accept more and more features, and features are great, but IMO what we do need to focus on right now is: partial refactoring, better testing suit, bugfixes, benchmarks. New features also mostly require all of these, since adding new code means also a bigger maintaining burden.
If we do though proceed through the route of adding more side-features (like this PR and overall color picking; screencast, etc) and becoming more of a all-in-one tool, I think these functionality must at least be separated in different crates, like we do with `waysip`.
I'd like to discuss the overall view of wayshot goals with other members, so I'll send the message in discord a bit later, to not pollute this thread.

Gigas002:
~~Ah, also I think it'd be better to use color as command rather than flag, e.g:~~

```
wayshot color --format rgb
```

Ugh, no, that would break existing args (FILE)

0xsamalt:
> @0xsamalt by the way, I think you should also update the document.

I'll push the documentation updates shortly to finalize the flag options!

Regarding the bigger picture, @Gigas002 makes a fair point about the UNIX philosophy versus an all-in-one approach. I'll keep an eye out for the discussion on Discord. Once the team comes to a consensus on the feature's alignment with wayshot's goals, just let me know how you'd like to proceed with this PR.

Shinyzenith:
> I'd like to discuss the overall view of wayshot goals with other members, so I'll send the message in discord a bit later, to not pollute this thread.

I largely agree with you. The goal of wayshot is to be a unified API over wayland protocols that you can call easily and just forget about the details. We are accepting too many features at once. UNIX like style is preferred but I am not a BDFL so we can all discuss how we want to shape things.

Gigas002:
After giving some thoughts on this PR, I think I'm not against having such functionality - if we have it hidden with feature and it has clear separation code-wise. 
I am going to work on overall cli code refactoring in upcoming days, and I think afterward it'll be much easier and cleaner for us to add new features.

Decodetalkers:
seems you need to rebase now. Sorry, we have merged an important refactor

Gigas002:
Shouldn't be too hard to rebase, hope these notes will help you with it:
- cli will most likely be good as is, just move the enum to `color_picker.rs`
- new func in `wayshot.rs` should also be moved into `color_picker.rs`
- add new options resolvement into `settings.rs`
- call new logic in `wayshot.rs`
- docs rebasing shouldn't be too hard, nothing to add here

Gigas002:
@Decodetalkers 

Rebased this on `main` and co-authored @0xsamalt. Can you please check when you have time?

Decodetalkers (approved):
LGTM, all thinks looks good, approved

As soon as now the part is split to a feature, I think it will be find

PR #158, opened 2025-03-27
feat: Add Notification for Screenshots in Wayshot

noobed-max:
Added a notification feature to notify users  when a screenshot is taken.

PR #306, opened 2026-03-03
refactoring: complete cli overhaul

Gigas002:
This PR implements major refactoring of `wayshot` cli tool. No changes made to `libwayshot`, `examples`, etc. 
No logic and breaking changes either.

Closes: #301 

Since PR is a bit huge, I tried to separate each task I wanted to achieve with this refactoring as separate commit, so it probably would be easier to review commit-by-commit. Especially since overall diff won't show easy to understand comparison because all files have their stuff moved here and there.

In details, this PR introduces following changes:

1. Separates `clipboard`-related code into `clipboard.rs`
2. Adds `clipboard` feature (concern about it was raised in #48). It simplifies dependency-tree for non-default-features builders and eliminates `unsafe` code in `wayshot` for such users (since you'll have to deal with `wl-copy` or similar tools by yourself)
3. Separates `color picker`-related code into `color_picker.rs`
4. Adds `color_picker` feature. It doesn't really depend on anything ATM, but since I guess it's theoretically possible that it'll add some burden in future I decided that it's not a bad idea to separate it right now while we can
5. Separates `logger` initialization into `logger.rs`
6. Adds `logger` feature (looking at `silent` option made me think about suppressing the logs)
7. Adds `selector` feature. Likewise `clipboard` can use 3rd party tools like `wl-copy`, that would be great to optionally support `slurp`/`waysip` (not `libwaysip`) like tools for users who don't need built-in selector. That would also eliminate source of errors, theoretically coming from `libwaysip`, simplifying bugs identification (still, for power users who build without features though). Anyway, that's plan for future, right now the feature just enables/disables built-in `libwaysip`
8. Adds `jxl` feature. Since `jxl` encoding requires native `libjxl` dependency this is a good idea, IMO. Also, `jxl` encoding without hardware acceleration sucks, so users without hw that supports it won't even need that feature and dependency at all
9. Adds `avif` feature. Doesn't require native dependencies, but preformance-wise concern is the same as `jxl`. And still adds some build-time deps
10. Separates `notification`-related code into `notification.rs`
11. Adds `notification` feature
12. Adds `no-notification` alias to `silent` option to better match config naming (`notification`)
13. Simplifies cli/config options and commands resolvement and separates it into `settings.rs`
14. Slightly simplifies some `utils.rs` funcs
15. Adds some minor explanation comments and expands existing ones, so new contributors will better understand what's going on and where to add stuff
16. Updates docs (no surprise)

Number of build crates drops from 372 to 171 when using `--no-default-features`, and resulting binary size drops from 11Mb to 2Mb.

For future plans that I want to implement next, but didn't cover here:
- add `freeze` cli option
- add `delay` option (will supersede #139)
- bump dependencies at once
- consider adding preview image to notification (was discussed in #158)
- additionally support 3rd party tools for geometry handling (`selector` feature related; we already had this behavior before `libwaysip` integration)
- improve test coverage and CI
- consider adding benchmarks (#45, #103)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,16 @@ cargo build --no-default-features --features clipboard,logger,notifications
 | `notifications`| Desktop notifications after each capture              | notify-rust           |
 | `selector`     | `--geometry` flag, interactive region selection       | libwaysip             |
 
+## Clipboard without the built-in feature
+
+If you build without the `clipboard` feature, you can still pipe stdout to `wl-copy`:
+
+```bash
+wayshot - | wl-copy
+```
+
+Alternatively, set `stdout = true` in your config file to always write to stdout, then pipe as usual.
+
 # Installation
 
 ## AUR:


Shinyzenith:
Hi initial thoughts -- looks mostly good imo but your build ci is failing so can you please check that?

Gigas002:
Sure, will fix it. 
It's rust fmt that's not happy with latest commits, forgot to run it before committing. Build itself works fine.

Decodetalkers:
Then we need to modify the ci.yml, make it tests all the features, thanks for your great work! it becomes better

Gigas002:
I was thinking about CI fixes for a while (was planning to do it in different PR, sorry I still don't have enough time for write a proper explanation message for this PR).
So, what I was thinking about, is that we should probably run tests only on libwayshot, I'm not sure there's a point in testing the cli itself, but I'm unsure.
Anyway, I was planning on improving the CI and tests next, we can discuss it in corresponding PR when I'm done

Gigas002:
Oof, I think I wrote every little thought I had about this PR in description. This should be in a good state for review now.

PR #307, opened 2026-03-09
fix: shm formats can contains many

Decodetalkers:
issue:https://github.com/waycrate/xdg-desktop-portal-luminous/issues/168

the origin logic only allow one format, make this error happen


Decodetalkers:
@abcfy2 can you build the target your self, with the libwayshot in this pr, to check if it is fixed?

Gigas002 (approved):
@Decodetalkers 
Thanks, LGTM!

PR #311, opened 2026-03-13
fix: licensing issues

Gigas002:
As discussed with @Shinyzenith on discord, we have `waymirror-egl` and `wayshot` dependent on strong copyleft licenses and have to either remove features, or change licenses to comply.
The `libwayshot` is unaffected.
Also added `deny.toml` config and corresponding job.
Locally checkable via: 

```bash
cargo deny check licenses
```

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -1,0 +1,17 @@
+name: cargo-deny
+
+on: [push, pull_request]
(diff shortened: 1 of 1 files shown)


Decodetalkers:
I think it's fine, as if @Shinyzenith approves, it should be merged

Issue #234, opened 2025-12-06
About wayshot 1.4.0 prepare


Decodetalkers:
We are thinking about publish a new version for wayshot. Because it has changed a lot, I think next version should be 1.4.0. Before this version is published, we need:
 
* An auto publish ci
* add CHANGELOG.md

roland-5:
Is [Wayshot 1.4.1 & libwayshot 0.5.0](https://github.com/waycrate/wayshot/releases/tag/v1.4.1) considered stable? I see that Arch Linux AUR packages are not updated.

Decodetalkers:
> Is [Wayshot 1.4.1 & libwayshot 0.5.0](https://github.com/waycrate/wayshot/releases/tag/v1.4.1) considered stable? I see that Arch Linux AUR packages are not updated.

Not very stable yet.. As you have seen the api still need to be tidyup. But wayshot is not my part.. 

We will publish a new one today. Now it is more stable.

Emm.. I do not know who is responsible for the package on aur...

Decodetalkers:
Although I also want to tidy up the api.. seems now it already has the full api. So I think now it is stable.After @Shinyzenith come back, that time we will consider starting to tidy up the api.

I think now it is stable..

roland-5:
Thank you for answer. `wayshot` maintainer is `eclairevoyant` and for `wayshot-bin` is `Shinyzenith`, so Shinyzenith will probably update package when he comes back.

Shinyzenith:
I do not maintain aur binaries anymore, infact I do not intend to distribute x86 binaries anymore.

Shinyzenith:
If anyone wants, they can take over.

Anshi321:
Hi, @Decodetalkers  I’d like to help with preparing the 1.4.0 release.
I can work on adding the CHANGELOG.md if that’s still needed.
Let me know and I’ll wait for confirmation before opening a PR.

Decodetalkers:
> Hi, [@Decodetalkers](https://github.com/Decodetalkers) I’d like to help with preparing the 1.4.0 release. I can work on adding the CHANGELOG.md if that’s still needed. Let me know and I’ll wait for confirmation before opening a PR.

but this issue is already over. I think. I forgot to close it. but still remained problems. The auto publish ci seems broken. https://github.com/waycrate/wayshot/blob/main/.github/workflows/build.yml#L131-L136 it should can read the CHANGELOG.md and auto publish, but every time it fails

[...]

PR #315, opened 2026-03-18
refactor: refactor CI by separating on several files

Gigas002:
Mostly moving stuff around, but minor changes included:
- build three patterns: `all-features`, `default-features`, `minimal` (with just `png` feature)
- separate coverage reports for `libwayshot` and `wayshot`

Also, left unchanged, but in #234 @Shinyzenith mentioned that probably we don't need to distribute binaries with release, maybe we should remove this part of job? If not, we should probably add `all-features` to it? Or distribute several flavors? Or just the `default`? I'm not sure what'd be best here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build Wayshot
+name: Build
 
 on: [push, pull_request]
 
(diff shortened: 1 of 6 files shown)


Decodetalkers (approved):
Seems good for me! I have thought this ci is too long long ago

PR #132, opened 2025-03-08
Implemented completions for various shell

Rishik-Y:
Implements auto-completion for various shells such as Bash, fish, zsh, powershell, elvish etc

Example:
```
wayshot - [TAB]
-s                      --extension
-c                      --format
-l                      --output-format
-o                      --encoding
-h                      --list-outputs
-V                      --output
--clipboard             --choose-output
--log-level             --generate-completions
--slurp                 --help
--cursor                --version
```

or 

```
wayshot --output-format 
jpg   png   ppm   qoi   webp  
```

PR #319, opened 2026-03-25
feature: shell completions

Gigas002:
Add shell completions for various shells, including `nushell` via `clap_complete_nushell` crate. Includes `completions` feature enabled by default.
Comparing to #132, this PR doesn't include pre-built completions, as these should be generated on user side, depending on feature set selected.

Closes: #100


Decodetalkers:
Seems good!

PR #320, opened 2026-03-26
refactor: hide egl code behind feature in libwayshot

Gigas002:
This PR hides `egl`-related code in `libwayshot` behind feature (enabled by default).
This may seem useless at the moment, but this is part of further refactoring before implementing optional `vulkan` support. I've been testing and using it in my fork for ~2 weeks and things are looking good.

**Breaking changes**

While this is not breaking for most users, `default-features = false` users will lose `egl` support. Migration step is to add feature, e.g. `libwayshot = { version = "...", default-features = false, features = ["png", "egl"] }`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
                     - name: default
                       cargo_args: --workspace --release
                     - name: minimal
-                      cargo_args: --workspace --release --no-default-features --features "egl,png"
+                      cargo_args: --workspace --release --no-default-features
 
         name: ${{ matrix.name }}
         runs-on: ubuntu-latest


Gigas002:
Hmm, probably we don't even need it as default feature in cli...

Decodetalkers (approved):

Decodetalkers:
oh.. sorry, this is just a draft. I should not approve it now

Gigas002:
Yeah, my bad. I've opened it yesterday, but now I thought that maybe we don't need that feature on by default in cli, since we don't use egl on it to capture screenshot anyway

Gigas002:
@Decodetalkers 
Sorry, I've removed the `egl` feat from `wayshot` cli. Can you please re-review when you're free? 
Thank you!

Decodetalkers (approved):

Issue #327, opened 2026-03-30
Corrupted screenshots (banding / stale regions) due to premature frame consumption after FrameState::Finished


HiFiveJazz:
Hi! I’ve been investigating an issue with `libwayshot` where screenshots occasionally contain corrupted regions (banding / partially stale frames), particularly on multi-monitor setups and higher refresh rate displays. This discussion originally came from a [woomer issue](https://github.com/coffeeispower/woomer/issues/32) but after isolating the behavior it appears reproducible directly with `libwayshot`. 

I cannot yet confirm whether this started after a specific compositor change, but I began noticing it on Hyprland 0.54.3.

If I run the following code using the latest `wayshot` repo:

```rust
// main.rs
use libwayshot::WayshotConnection;


fn main() {
    let wayshot = WayshotConnection::new()
        .expect("failed to connect to wayshot");

    wayshot
        .screenshot_all(false)
        .expect("failed to capture screenshot")
        .save("screenshot.png")
        .expect("failed to save image");
}
```

I expect to get a clean screenshot of my desktop like so:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b7386eda-cfcc-4a36-8e69-4ee5750b5b68" />

However, screenshots consistently looked like this

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/695db089-4ee2-448c-8ff5-7455135e9ec2" />

Screenshots often had this horizontal banding whose region varied from a previous frame. This banding only occurred on my primary monitor, 1920x1080@360Hz but never on my secondary monitor, 1920x1200@60Hz monitor.

Interestingly, using [grim-rs](https://github.com/shikoucore/grim-rs), a recently deprecated wayland screenshot tool did not give me these banding issues.

After tinkering with the codebase for a couple hours I have some key findings:

- The issue occurs even when forcing the wlr-screencopy path (not specific to ext-image-copy)
- The issue is not related to color format conversion (RGBA/BGRA)
- The issue is not reproducible with grim
- The issue is timing/synchronization related 

Critically, I observed that within `lib.rs` returning `FrameState::Finished` can produce corrupted output.

However, the issue disappears when adding a delay into the code like so

```rust
 //lib.rs, line 877
        loop {
            // Basically reads, if frame state is not None then...
            if let Some(state) = state.state {
                match state {
                    FrameState::Failed => {
                        tracing::error!("Frame copy failed");
                        return Err(Error::FramecopyFailed);
[...]

PR #328, opened 2026-03-30
fix: ensure frame synchronization after FrameState::Finished to preve…

HiFiveJazz:
Returning immediately after `FrameState::Finished` can result in corrupted
captures (e.g., banding or stale regions), especially on high refresh rate
displays.

This change adds a Wayland event queue synchronization step (`roundtrip`)
after `Finished` before returning the captured buffer.

Without this, the SHM buffer may be consumed before all relevant events
have been fully processed. Adding the synchronization ensures the frame
is fully ready before use.

This behavior was reproducible on Hyprland (144 Hz), and the fix eliminates
the issue consistently.

Tested by:
- comparing against grim output (which does not exhibit the issue)
- verifying removal of banding artifacts across repeated runs. 

For more information please see #327

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -876,8 +876,8 @@ impl WayshotConnection {
         // On copy the Ready / Failed events are fired by the frame object, so here we check for them.
         loop {
             // Basically reads, if frame state is not None then...
-            if let Some(state) = state.state {
-                match state {
+            if let Some(frame_state) = state.state {
+                match frame_state {
                     FrameState::Failed => {
                         tracing::error!("Frame copy failed");
                         return Err(Error::FramecopyFailed);
@@ -887,6 +887,7 @@ impl WayshotConnection {
                         return Err(Error::FramecopyFailedWithReason(reason));
                     }
                     FrameState::Finished => {
+                        event_queue.roundtrip(&mut state)?;
                         tracing::trace!("Frame copy finished");
                         return Ok(DMAFrameGuard {
                             buffer: dmabuf_wlbuf,
@@ -949,8 +950,8 @@ impl WayshotConnection {
         // On copy the Ready / Failed events are fired by the frame object, so here we check for them.
         loop {
             // Basically reads, if frame state is not None then...
-            if let Some(state) = state.state {
-                match state {
+            if let Some(frame_state) = state.state {
+                match frame_state {
                     FrameState::Failed => {
                         tracing::error!("Frame copy failed");
                         return Err(Error::FramecopyFailed);
@@ -960,8 +961,8 @@ impl WayshotConnection {
                         return Err(Error::FramecopyFailedWithReason(reason));
                     }
                     FrameState::Finished => {
+                        event_queue.roundtrip(&mut state)?;
                         tracing::trace!("Frame copy finished");
-
                         return Ok(DMAFrameGuard {
                             buffer: dmabuf_wlbuf,
                         });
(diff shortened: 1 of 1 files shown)


HiFiveJazz:
@Gigas002 Thank you for the push to investigate and contribute :)

Gigas002:
Hmm, I've just encountered this bug again. I guess it's not completely fixed yet...

Gigas002:
@Decodetalkers @HiFiveJazz 
I feel like there's a race condition in `freeze` maybe? Below are the screenshots and how I reproduced it.

To screenshot a fullscreen app/game with custom cursor, I usually:
1. Press on screenshot to freeze it
2. Change hyprland workspace to use cursor for selector tool
3. Select geometry and get the screenshot flow completed

So, here I tried screenshoting the fullscreen game using this flow:

image

This is a working example, nothing stands out. However, sometimes, this happens:

image

For some reason, part of screenshot isn't aquired from freezed image, but from actual frame behind the frozen one (plus changed color).

Decodetalkers:
> @Decodetalkers @HiFiveJazz 
> I feel like there's a race condition in `freeze` maybe? Below are the screenshots and how I reproduced it.
> 
> To screenshot a fullscreen app/game with custom cursor, I usually:
> 1. Press on screenshot to freeze it
> 2. Change hyprland workspace to use cursor for selector tool
> 3. Select geometry and get the screenshot flow completed
> 
> So, here I tried screenshoting the fullscreen game using this flow:
> 
> image
> 
> This is a working example, nothing stands out. However, sometimes, this happens:
> 
> image
> 
> For some reason, part of screenshot isn't aquired from freezed image, but from actual frame behind the frozen one (plus changed color). 

Maybe we need help from hyperland.ha.., I do not know. When we receive finish signal from the wm, the screenshot should have finished..

HiFiveJazz:
> > @Decodetalkers @HiFiveJazz
> > I feel like there's a race condition in `freeze` maybe? Below are the screenshots and how I reproduced it.
> > To screenshot a fullscreen app/game with custom cursor, I usually:
> > 
> > 1. Press on screenshot to freeze it
> > 2. Change hyprland workspace to use cursor for selector tool
> > 3. Select geometry and get the screenshot flow completed
> > 
> > So, here I tried screenshoting the fullscreen game using this flow:
> > image
> > This is a working example, nothing stands out. However, sometimes, this happens:
> > image
> > For some reason, part of screenshot isn't aquired from freezed image, but from actual frame behind the frozen one (plus changed color).
> 
> Maybe we need help from hyperland.ha.., I do not know. When we receive finish signal from the wm, the screenshot should have finished..

Hmm, interesting, is the different color banding happening inconsistently like the original bug? What I mean to say is it the bug having the entire object behind the fullscreen be a different color or sometimes inconsistent bands of the object behind the fullscreen be a different color like the original?

Would agree that it's definitely some sort of a race condition. Like @Decodetalkers  I'm also skeptical because thorough examination of the code didn't seem like anything would result in a race condition, or getting the original image one frame too early. 

If placing another wayland roundtrip or timing delay in the code solves the issue then we found where it lives but even in this PR we merged, it's possible Hyprland is presenting an image that is not ready and saying it is ready too early, I couldn't tell if the issue was in libwayshot's implementation or the wayland protocol on my Hyprland system, I only know it exists somewhere within that pipeline.

I think it was for that reason that mutexes were use a lot in some other wayland screenshotting tools but I would have to do a deeper dive on the wayland protocol.

Assuming it exists in `libwayshot`, it's likely to be before `FrameState::Finished` is being consumed, I can see it being in `freeze` but it's also probable we may need to get Hyprland involved if we don't find any race conditions in `libwayshot`.

Gigas002:
Yeah, I also agree that we may need to consult with hyprland guys just to confirm

Created discussion in hyprland issue tracker: https://github.com/hyprwm/Hyprland/discussions/13952

PR #352, opened 2026-04-20
fix: make cargo doc pass without warns

Gigas002:
`cargo doc` job will help us catch warns in doc entries so docs.rs page stays happy


Gigas002:
Bruh it's easier to just use the same container `test-coverage` job uses to make `--all-features` pass

Decodetalkers (approved):

PR #354, opened 2026-04-21
fix: explicit MIME type handling for clipboard instead of autodetect

RustyCoderX:
noticed that clipboard MIME detection was only working properly for JPEG and PNG because it relied on autodetection. So I removed that and added explicit MIME types based on the image format instead.Also updated the clipboard function to take the encoding format so it knows exactly which MIME type to use, and passed that from where it’s called.This makes clipboard copy work consistently for formats like WebP, AVIF, JXL, etc., not just JPEG/PNG.

### DIFF
--- a/wayshot/src/clipboard.rs
+++ b/wayshot/src/clipboard.rs
@@ -4,6 +4,31 @@ use eyre::Result;
 use rustix::runtime::{self, Fork};
 use wl_clipboard_rs::copy::{MimeType, Options, Source};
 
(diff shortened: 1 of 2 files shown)


Decodetalkers (approved):

Decodetalkers:
@Gigas002 seems reasonable for me, may you also take a look?

Gigas002 (approved):
LGTM! Thank you for the fix!

PR #368, opened 2026-05-16
feature: add configuration for notifications

### LINKED ISSUE #334: feature request: consider providing configuration for notifications
As discussed in #332, some users would like to configure notifications a bit more thoroughly, and `notify-rust` provides such settings, so it shouldn't be complicated to implement for config.

Gigas002:
Closes: #334 

Actual notification features differs depending on user's notification daemon, e.g. most lightweight daemons like `dunst` or `mako` doesn't support sound playback, so the parameter doesn't have any effect.

Examples:

image

image

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -96,10 +96,6 @@ effort = 7
 # -1 = never expire, 0 = server default, positive = milliseconds.
 # timeout_ms = 5000
 
-# Urgency level: "low", "normal", or "critical".
-# Critical notifications do not time out on most servers.
-# urgency = "normal"
-
 # Application name shown in the notification.
 # app_name = "wayshot"
 
--- a/config.toml
+++ b/config.toml
@@ -88,3 +88,33 @@ effort = 7
 # The command is run exactly as provided with no arguments appended.
 # If not set, opens the screenshot directory with xdg-open.
 # action = "xdg-open PATH"
+
+# Icon name shown in the notification (uses icon theme lookup).
+# icon = "camera-photo"
+
+# How long the notification is visible, in milliseconds.
+# -1 = never expire, 0 = server default, positive = milliseconds.
+# timeout_ms = 5000
+
+# Urgency level: "low", "normal", or "critical".
+# Critical notifications do not time out on most servers.
+# urgency = "normal"
+
+# Application name shown in the notification.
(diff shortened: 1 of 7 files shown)


config.toml
```diff
@@ -88,3 +88,33 @@ effort = 7
...
+# Icon name shown in the notification (uses icon theme lookup).
+# icon = "camera-photo"
+
+# How long the notification is visible, in milliseconds.
+# -1 = never expire, 0 = server default, positive = milliseconds.
+# timeout_ms = 5000
+
+# Urgency level: "low", "normal", or "critical".
+# Critical notifications do not time out on most servers.
+# urgency = "normal"
```
Decodetalkers: I think the urgency should not be in config 

What I have thought is that urgency should be split into two cases. For example, if it succeeds, it should be normal, if it fails, it should be critical. And it should be defined in arguments, not default in configuration.

Gigas002: Yeah, sound reasonable

Gigas002: Fixed

wayshot/src/notification.rs
```diff
@@ -60,18 +59,53 @@ pub fn send_success(
...
+        let _ = notification.show();
     }
 }
 
-pub fn send_failure(error: &Error) {
-    let _ = Notification::new()
-        .summary("Screenshot Failed")
+pub fn send_failure(error: &Error, config: &NotificationConfig) {
+    let _ = build_base_notification(config, false)
+        .urgency(Urgency::Critical)
```
Decodetalkers, line 72: I do not think failure should always be critical.. I think it would be better to have an argument to control the action of failure. 

Emm.. when will the screenshot be that important, if it fails, the notification should be marked as critical?.

Decodetalkers:
Wait.. maybe we should just control the action of failure? Or maybe always being critical will be enough? I want to hear more opinions 

This part maybe good. We can ship it. But I want let another person also decide it

Decodetalkers:
+1

Shinyzenith:
I think I'm happy with this and we can merge, took an overall glance at the code, if you're happy with it @Decodetalkers merge it @Gigas002

Decodetalkers (approved):

PR #370, opened 2026-05-26
chore: update changelog for 1.5.0 release

Gigas002:
There's no real tag yet and date _may_ be not today, but for now set to these values if we'll manage to do all today

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -4,3 +4,4 @@ paeth = "paeth"
 # Git short hashes are not dictionary words (e.g. `ba` in e9ba46b).
 [default.extend-identifiers]
 e9ba46b = "e9ba46b"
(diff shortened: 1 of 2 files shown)


Decodetalkers (approved):

Decodetalkers:
LGTM, then I will make a tag for it, it will be pulished to crate.io sooner

Gigas002:
@Decodetalkers 
Thanks and sorry for wait

Decodetalkers:
And should we also bump the verison of wayshot? I think it also induced a lot of features, and it should also rely on new libwayshot. should it also be a breaking change?

Decodetalkers:
ohh... I see. I will do that, 1.5.0

PR #381, opened 2026-07-02
chore(screenshot)!: use identifier to select toplevel

### LINKED ISSUE #379: Feature: be able to capture a window by its id
Currently it's only possible to capture a window by its app_id and title (`--toplevel`). But what if there are multiple windows with the same app id/class and title?

So there would be an option to capture a window by its unique id (`foreign_toplevel_identifier` on Sway)

I love this software please implement this 🙏

Decodetalkers:
resolve: https://github.com/waycrate/wayshot/issues/379

it is a breaking change, the origin logic is use id+title to choose a toplevel to shot, but the information is very verbose. Before I thought identifier is very random, so I did not use it to find the toplevel, but we can also print the identifier with list-toplevels, so I also modified that part to show the identifier information

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -61,7 +61,7 @@ pub struct Cli {
     #[arg(long)]
     pub list_outputs_info: bool,
 
-    /// List all active toplevel windows.
+    /// List all active toplevel windows. you will the information about id+title+identifier
     #[arg(long, alias = "list-windows")]
     pub list_toplevels: bool,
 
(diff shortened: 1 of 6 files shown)


wayshot/src/cli.rs
```diff
@@ -75,21 +61,14 @@ pub struct Cli {
     #[arg(long)]
     pub list_outputs_info: bool,
 
-    /// List all active toplevel windows.
+    /// List all active toplevel windows. you will the information about id+title+identifier
```
Gigas002:
```suggestion
    /// List all active toplevel windows. You will see the information about id+title+identifier
```

Gigas002 (changes requested):
@Decodetalkers 
Thanks for the fix! I think this is a good change, and makes much easier to use.
Could you please also update the docs and changelog with minor notes for migration steps since this is a breaking change?

Gigas002:
@Decodetalkers 
Ok, LGTM, thank you!
I guess we can update changelog later, closer to next release so it's fine

Gigas002 (approved):

PR #387, opened 2026-07-08
feature: list outputs and toplevels as json

Diax170:
Add `--list-outputs-json` and `--list-toplevels-json` commands for easier scripting

This PR is mostly copy+paste garbage code but hey, it works


Diax170:
> ❌ Patch coverage is `0%` with `45 lines` in your changes missing coverage. Please review.

what does that mean

Decodetalkers:
@Gigas002 what do you think about this pr?

libwayshot/src/lib.rs
```diff
@@ -360,6 +362,52 @@ impl WayshotConnection {
         }
     }
 
+    /// print the displays' info in JSON format
+    pub fn print_displays_info_json(&self) {
```
Decodetalkers: you should mark the structure with serde derive, then you do not need such line block

Decodetalkers (changes requested):

libwayshot/src/lib.rs
Decodetalkers: for example
https://github.com/waycrate/wayshot/blob/main/wayshot/src/config.rs#L7-L15

Decodetalkers: ok, you can also create a wrapper structure, from OutputInfo, this will also reduce the use of serde::Value

libwayshot/src/region.rs
```diff
@@ -57,6 +58,13 @@ impl TopLevel {
             self.app_id, self.title, self.identifier
         )
     }
+    pub fn id_title_identifier_json(&self) -> Value {
```
Decodetalkers: this part is acceptable

libwayshot/src/lib.rs
Decodetalkers: ok, maybe acceptable.. we do not need an structure here

Decodetalkers:
How to regret the request changes.. ok.. if Gigas approve it , I will also approve it

libwayshot/src/lib.rs
Decodetalkers: but must we add the logic in libwayshot? should not we add it in wayshot...But we already have print_display_info here.. Maybe it is meaningful

Diax170: > you should mark the structure with serde derive, then you do not need such line block

I'll have a look at it

Decodetalkers: > > you should mark the structure with serde derive, then you do not need such line block
> 
> I'll have a look at it

Sorry, I regret it .. seems we do not need a serde here.. it is just used to print something, Creating an extra structure for it maybe meaningless I am afraid.. I want to regret that request changes.. But I cannot find a button to cancel it

Diax170: It's OK

I have the ability to resolve this discussion, let me do it

Also let me know if you have any questions

Diax170: > but must we add the logic in libwayshot? should not we add it in wayshot...But we already have print_display_info here.. Maybe it is meaningful

idk what to think about it to be honest

I thought it made more sense as other wrappers are also here but I can move it out of libwayshot if you want

Gigas002 (changes requested):
@Diax170 
Thanks for contributing! 
There are several things I'd like to rework:
 
1. Please keep json out of `libwayshot`: no `serde_json` dep, no `print_*_json` funcs. The lib already exposes outputs/toplevels and ser/de stuff must be done on client side (in `wayshot` or any other crate users if they are willing to implement it in any other form)
2. Please prefer using `#[derive(Serialize)]` listing structs (e.g. in `wayshot/src/listing.rs`) over `json!` approach. Strings are error-prone and a hassle to deal with overall
3. Please update the documentation for CLI users with this new feature

Happy to re-review after that shape!

Diax170:
Alright, I somehow managed to remove serde_json from libwayshot without using any sorts of ai, and updated the docs so far. Will work on Serialize soon

Also, I'm using multiple commits to keep things more organized; you might want to squash them when merging

Diax170:
~Also, I noticed that the `libwayshot::region` imports are a bit scattered around, should I only import `libwayshot::region` and use `region::Position`, `region::Size` etc. or import all things directly from `libwayshot::region` into the scope, like this `use libwayshot::region::{LogicalRegion, Position, Region, Size};`?~

Never mind, after I added the Serialize structs those imports are no longer necessary

Diax170:
alr I did something, you can reivew now

Gigas002 (approved):
LGTM, thanks!

@Decodetalkers 
Can you check this out as well please?

Diax170:
why did the nix build also fail here? on my machine it doesn't build either. i am now rebuilding on different revisions to see where the issue is

Diax170:
works on 892512c1 and fails on cc1ed85c so that probably got messed up in #392

edit: fix typo

Gigas002:
Should be fixed after: #396

Diax170:
sorry i accidentally re-requested review :(

Cargo.toml
```diff
@@ -23,3 +23,4 @@ image = { version = "0.25", default-features = false }
 wayland-client = "0.31"
 wayland-protocols = "0.32"
 gl = "0.14"
+serde_json = "1"
```
Gigas002: Since we're not using json capabilities in `libwayshot` this can be moved into `wayshot`'s `Cargo.toml`. No need for workspace-level dep for now

Gigas002 (changes requested):

Cargo.toml
Gigas002: (I overlooked this on previous review)

Decodetalkers:
@Gigas002 by the way, should we also remove the print information logic in libwayshot and move it to wayshot? it is quite useless in library crate..

Gigas002:
@Decodetalkers 
Yeah, that's a good idea

Cargo.toml
Diax170: thanks for pointing it out, I forgot to remove it earlier

Diax170: how's this?

Diax170:
also, I'm new to convenitonal commits and curious why are you using "chore:" instead of "refactor:"?

Cargo.toml
Gigas002: Great, thanks :)

Gigas002 (approved):

Decodetalkers (approved):

Gigas002:
@Diax170 
Thanks, merged!
Regarding commit/PR names, I'd say that's more or less up to contributor taste. If you can understand what the contrib is about by looking at the commit name that's enough, we're not _too strict_ about these rules. 
Sometimes I, by mistake, mark `refactor` commits as `fix` as it aligns in my head as `architecture fix` lol.

PR #388, opened 2026-07-09
nix: Remove invalid changelog from flake.nix

Diax170:
Since this is the Git version, there aren't really changelogs. Previous changelog field was pointing to an invalid release anyway

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -39,7 +39,6 @@
             meta = {
               description = "Screenshot crate for wlroots based compositors implementing the zwlr_screencopy_v1 protocol.";
               homepage = "https://crates.io/crates/wayshot";
-              changelog = "https://github.com/waycrate/wayshot/releases/tag/v${finalAttrs.version}";
               license = with lib.licenses; [
                 bsd2
                 gpl3Only


flake.nix
```diff
@@ -39,7 +39,6 @@
             meta = {
               description = "Screenshot crate for wlroots based compositors implementing the zwlr_screencopy_v1 protocol.";
               homepage = "https://crates.io/crates/wayshot";
-              changelog = "https://github.com/waycrate/wayshot/releases/tag/v${finalAttrs.version}";
```
Gigas002, line 42: > Since this is the Git version

Shouldn't it point to main branch's changelog file then?

id3v1669: Can be either. 
meta is cosmetics anyway in this case, so can be removed or link `https://github.com/waycrate/wayshot/blob/main/CHANGELOG.md`
Whatever you prefer.

Gigas002: Removal feels fine then, less of a maintainance burden

Gigas002 (approved):
LGTM, thanks!

PR #390, opened 2026-07-13
fix: reuse all wayland registers, use refcell to reuse queue

### LINKED ISSUE #389: Memory leak on screencast
every second its usage increases by ~1mb
this is a huge memory leak

this was tested on hyprland, on sway it doesnt work and eventually just takes down the entire wayland session
usage seen on "resources" program

Decodetalkers:
and remove embedding region, move all the instance to WayshotScreencast, not create them every frame, this will let the memory stop being bigger

Fix: https://github.com/waycrate/wayshot/issues/389

### DIFF
--- a/libwayshot/Cargo.toml
+++ b/libwayshot/Cargo.toml
@@ -9,13 +9,14 @@ repository.workspace = true
 edition.workspace = true
 
 [features]
-default = ["png", "egl"]
+default = ["png", "egl", "screencast"]
 png = ["image/png"]
 jpeg = ["image/jpeg"]
 qoi = ["image/qoi"]
 webp = ["image/webp"]
 avif = ["image/avif"]
 egl = ["dep:gl", "dep:r-egl-wayland"]
+screencast = []
 
 [dependencies]
 tracing.workspace = true
--- a/libwayshot/examples/waymirror.rs
+++ b/libwayshot/examples/waymirror.rs
@@ -47,7 +47,7 @@ fn main() {
         .try_init_dmabuf(WayshotTarget::Screen(output.clone()))
         .expect("Cannot find a drm");
     let cast = wayshot
-        .create_screencast_with_dmabuf(WayshotTarget::Screen(output), true, None)
+        .create_screencast_with_dmabuf(WayshotTarget::Screen(output), true)
         .unwrap();
 
     let view_porter = globals
@@ -56,7 +56,6 @@ fn main() {
     let viewport = view_porter.get_viewport(&base_surface, &qhandle, ());
 
     let mut state = State {
-        wayshot,
         running: true,
         base_surface,
         viewport,
@@ -83,7 +82,6 @@ fn main() {
 }
 
 struct State {
-    wayshot: WayshotConnection,
     running: bool,
     base_surface: wl_surface::WlSurface,
     viewport: wp_viewport::WpViewport,
@@ -96,7 +94,7 @@ struct State {
 
 impl State {
     fn refresh_surface(&mut self) -> libwayshot::Result<()> {
-        self.wayshot.screencast(&mut self.cast)?;
+        self.cast.screencast()?;
 
         self.cast_size = self.cast.current_size();
         self.base_surface.attach(Some(self.cast.buffer()), 0, 0);
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -202,7 +202,7 @@ impl WayshotConnection {
     }
 
     // NOTE: this function says if we need to try receive the gdm information from ext-image-copy
-    fn need_try_find_gbm(&self) -> bool {
+    pub(crate) fn need_try_find_gbm(&self) -> bool {
         !self.has_gbm() && self.find_dmabuf
     }
     ///Create a WayshotConnection struct having DMA-BUF support
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -112,6 +112,7 @@ pub struct WayshotConnection {
     toplevel_capture_support: bool,
     image_copy_support: bool,
     find_dmabuf: bool,
+    registers: WayshotRegisters,
 }
 
 pub enum WayshotFrame {
@@ -155,23 +156,22 @@ impl WayshotTarget {
     }
 }
 
-fn check_toplevel_protocols(globals: &GlobalList, conn: &Connection) -> Result<()> {
-    let event_queue = conn.new_event_queue::<CaptureFrameState>();
-    let qh = event_queue.handle();
-
-    // Bind managers
-    globals.bind::<ExtImageCopyCaptureManagerV1, _, _>(&qh, 1..=1, ())?;
-    globals.bind::<ExtForeignToplevelImageCaptureSourceManagerV1, _, _>(&qh, 1..=1, ())?;
-    Ok(())
-}
-
-fn check_ext_image_copy_protocols(globals: &GlobalList, conn: &Connection) -> Result<()> {
-    let event_queue = conn.new_event_queue::<CaptureFrameState>();
-    let qh = event_queue.handle();
-
-    // Bind managers
-    globals.bind::<ExtImageCopyCaptureManagerV1, _, _>(&qh, 1..=1, ())?;
-    Ok(())
+#[derive(Debug)]
+pub(crate) struct WayshotRegisters {
+    pub layer_event_queue: EventQueue<LayerShellState>,
+    pub capture_event_queue: EventQueue<CaptureFrameState>,
+    pub output_event_queue: EventQueue<OutputCaptureState>,
+    pub zxdg_output_manager: ZxdgOutputManagerV1,
+    #[allow(unused)]
+    pub toplevel_list: Option<ExtForeignToplevelListV1>,
+    pub layer_shell: Option<ZwlrLayerShellV1>,
+    pub compositor: WlCompositor,
+    pub viewporter: Option<WpViewporter>,
+    pub shm: WlShm,
+    pub screencopy_manager: Option<ZwlrScreencopyManagerV1>,
+    pub image_copy_capture_manager: Option<ExtImageCopyCaptureManagerV1>,
+    pub output_image_management: Option<ExtOutputImageCaptureSourceManagerV1>,
+    pub toplevel_source_manager: Option<ExtForeignToplevelImageCaptureSourceManagerV1>,
 }
 
 impl WayshotConnection {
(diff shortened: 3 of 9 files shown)


Decodetalkers:
it still becomes bigger sometime..  though

Decodetalkers:
Have asked the developer of wayland-clients.. every time we bind a new one to globals, it will be storaged to the globallists, without release. So we should reuse them

Shinyzenith (dismissed):

Gigas002:
@Decodetalkers 
Can you please wait until I check this out? I hope to look at it until the weekend.

Decodetalkers:
Ok!

Decodetalkers:
You can add a change request here, I maybe click the button by mistake

Gigas002 (changes requested):
Ok! Will set as request changes for a while then :)

Gigas002:
@Decodetalkers

Thanks for the fix! I don't think we need all of these changes though. I can see two things being done here:

1. `WayshotScreenCast` now owns a persistent `EventQueue` and its own bound managers and `screencast()` moved onto `WayshotScreenCast` itself instead of being re-derived from `WayshotConnection` on every frame via `capture_target_frame_get_state`. This part is good as is, IMO, this should be a sufficent fix for #389.
2. `WayshotRegisters` part - it caches every global on `WayshotConnection` itself. This forces `&mut self` onto everything downstream in client side.

I don't think we need to implement `WayshotRegisters` bits... It introduces new allocations on what used to be zero-cost accessors. `get_all_outputs()` and `get_all_toplevels()` went from returning `&[T]` to `.clone()`-ing a `Vec<T>` on every call.
It also removes the ability to share a connection behind `&`. A screenshot is a read-only, idempotent query against compositor state. Any consumer holding `Arc<WayshotConnection>` for concurrent reads, or just calling two independent query methods without threading `&mut` through their whole call stack, now can't.
I also think that shared mutable event-queue state is a correctness risk, not just a style preference. `capture_event_queue` is now shared across screenshot, toplevel capture, etc; correctness depends on every code path fully draining/resetting state before returning. That's a subtler invariant to maintain and test than "make a queue, use it, drop it."

Additionally, while reviewing this I noticed that we can feature-gate screencast in `libwayshot` (`features = ["screencast"]`), so wayshot or other consumers who don't need to compile the screencast feat don't pull in `gbm`/`drm`, which today are unconditional deps. This will be breaking (for libwayshot users, not wayshot-side), since I'd prefer to see this feature off by default, but migration step is a one-liner so I think it's not a big deal.

Decodetalkers:
> @Decodetalkers
> 
> Thanks for the fix! I don't think we need all of these changes though. I can see two things being done here:
> 
> 1. `WayshotScreenCast` now owns a persistent `EventQueue` and its own bound managers and `screencast()` moved onto `WayshotScreenCast` itself instead of being re-derived from `WayshotConnection` on every frame via `capture_target_frame_get_state`. This part is good as is, IMO, this should be a sufficent fix for #389.
> 2. `WayshotRegisters` part - it caches every global on `WayshotConnection` itself. This forces `&mut self` onto everything downstream in client side.
> 
> I don't think we need to implement `WayshotRegisters` bits... It introduces new allocations on what used to be zero-cost accessors. `get_all_outputs()` and `get_all_toplevels()` went from returning `&[T]` to `.clone()`-ing a `Vec<T>` on every call.
> It also removes the ability to share a connection behind `&`. A screenshot is a read-only, idempotent query against compositor state. Any consumer holding `Arc<WayshotConnection>` for concurrent reads, or just calling two independent query methods without threading `&mut` through their whole call stack, now can't.
> I also think that shared mutable event-queue state is a correctness risk, not just a style preference. `capture_event_queue` is now shared across screenshot, toplevel capture, etc; correctness depends on every code path fully draining/resetting state before returning. That's a subtler invariant to maintain and test than "make a queue, use it, drop it."
> 
> Additionally, while reviewing this I noticed that we can feature-gate screencast in `libwayshot` (`features = ["screencast"]`), so wayshot or other consumers who don't need to compile the screencast feat don't pull in `gbm`/`drm`, which today are unconditional deps. This will be breaking (for libwayshot users, not wayshot-side), since I'd prefer to see this feature off by default, but migration step is a one-liner so I think it's not a big deal.

I have thought the same thing before. I am not that confident  if event_queue will cause unreleased memory, so I cached them too. I have asked the developers of wayland-client, they reply we would better to reuse the resources. 

Because now it is everything &mut, so the reference will always be cloned in order to be used. So I replaced them with vec

Decodetalkers:
Ok. If queue itself won't allocate any extra memory, I will return the style of before.. 

I just afraid that it also cannot be freed as soon as wayshot connection exists

this was the developer's reply

image

Gigas002:
I see. Let me check out the `wayland-client` a bit before changes...

libwayshot/Cargo.toml
```diff
@@ -9,13 +9,14 @@ repository.workspace = true
...
 [features]
-default = ["png", "egl"]
+default = ["png", "egl", "screencast"]
 png = ["image/png"]
 jpeg = ["image/jpeg"]
 qoi = ["image/qoi"]
 webp = ["image/webp"]
 avif = ["image/avif"]
 egl = ["dep:gl", "dep:r-egl-wayland"]
+screencast = []
```
Gigas002, line 19: We don't need either `egl` or `screencast` for screenshotting wayshot binary and `gbm`/`drm` seem to only be used for `screencast`, so we can gate them as well

```suggestion
egl = ["dep:gl", "dep:r-egl-wayland", "screencast"]
screencast = ["dep:gbm", "dep:drm"]
```

Gigas002 (changes requested):
@Decodetalkers 
I agree with rebinding globals per-capture leaks (`wl_shm`, `wl_compositor`, `zwlr_screencopy_manager_v1`, etc. have no `destroy` request). Binding once in `init_bindings()` fixes that already. Can we drop the `WayshotRegisters`/`&mut self`/`Vec`-clone refactor on top of it? An `EventQueue` isn't a protocol object, so caching it isn't needed to fix the leak, just the bind.

As for additional leak, I found out that per protocol, `frame`/`session`/`source` need an explicit `.destroy()`:

- [ext-image-copy-capture-v1.xml](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/ext-image-copy-capture/ext-image-copy-capture-v1.xml) — `frame.ready`/`frame.failed`: "the client must destroy the object"
- [wlr-screencopy-unstable-v1.xml](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-screencopy-unstable-v1.xml) — same for `zwlr_screencopy_frame_v1.ready`/`.failed`

None of these are called in `screencast.rs`, and `screencast()` creates a new `source`+`session`+`frame` on every captured frame. Same applies to `screencast_wlr()`. That matches a per-frame leak better than the manager-rebinding one. I suggest to create `source`/`session` once per `WayshotScreenCast`, destroy in `Drop`; only `frame` gets created/destroyed per capture, on every exit path.

Decodetalkers:
> @Decodetalkers 
> I agree with rebinding globals per-capture leaks (`wl_shm`, `wl_compositor`, `zwlr_screencopy_manager_v1`, etc. have no `destroy` request). Binding once in `init_bindings()` fixes that already. Can we drop the `WayshotRegisters`/`&mut self`/`Vec`-clone refactor on top of it? An `EventQueue` isn't a protocol object, so caching it isn't needed to fix the leak, just the bind.
> 
> As for additional leak, I found out that per protocol, `frame`/`session`/`source` need an explicit `.destroy()`:
> 
> - [ext-image-copy-capture-v1.xml](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/ext-image-copy-capture/ext-image-copy-capture-v1.xml) — `frame.ready`/`frame.failed`: "the client must destroy the object"
> - [wlr-screencopy-unstable-v1.xml](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-screencopy-unstable-v1.xml) — same for `zwlr_screencopy_frame_v1.ready`/`.failed`
> 
> None of these are called in `screencast.rs`, and `screencast()` creates a new `source`+`session`+`frame` on every captured frame. Same applies to `screencast_wlr()`. That matches a per-frame leak better than the manager-rebinding one. I suggest to create `source`/`session` once per `WayshotScreenCast`, destroy in `Drop`; only `frame` gets created/destroyed per capture, on every exit path.

There are not &mut now, and now I reuse the queues with refcell, will it be acceptable? And WayshotRegisters is just a place to storage the reusable objects, and not I returned the vec back to &[]

Gigas002:
Yeah I think it's good. But I still think that the `frame`/`session`/`source` destroy fix in `screencast.rs` is needed. Drop for `WayshotScreenCast` only destroys `shm_pool`/`buffer` right now.

Roughly what I mean, sketched against current `screencast()`/`Drop`:

```rust
pub struct WayshotScreenCast {
    // ...
    image_copy_source: Option<ExtImageCaptureSourceV1>,
    image_copy_session: Option<ExtImageCopyCaptureSessionV1>,
}

impl Drop for WayshotScreenCast {
    fn drop(&mut self) {
        if let Some(session) = self.image_copy_session.take() {
            session.destroy();
        }
        if let Some(source) = self.image_copy_source.take() {
            source.destroy();
        }
        if let Some(pool) = self.shm_pool.take() {
            pool.destroy()
        }
        self.buffer.destroy();
    }
}

pub fn screencast(&mut self) -> Result<()> {
    if self.target.is_screen() && self.image_copy_manager.is_none() {
        return self.screencast_wlr();
    }
    
    // ...

    // create source+session once, reuse them on every subsequent call
    if self.image_copy_session.is_none() {
        let source = /* same create_source(...) call as today */;
        let session = image_copy_manager.create_session(&source, options, &qh, ());
        self.image_copy_source = Some(source);
        self.image_copy_session = Some(session);
    }
    let frame = self.image_copy_session.as_ref().unwrap().create_frame(&qh, ());

    // only `frame` is per-call now, destroy it on every exit path, success or error
    let result = (|| -> Result<()> {
        // existing dispatch/format/capture loop body
        Ok(())
    })();
    frame.destroy();
    result
}
```

Same `frame.destroy()` on every exit path wrapping applies to `screencast_wlr()`, just without the source/session part.

Decodetalkers:
so do you mean we can reuse the session? The logic current just create a new session every time..

Decodetalkers:
Seems a frame is copied, I will receive a stop event.. ok, I wlll invoke destroy after the framed is captured

libwayshot/Cargo.toml
Decodetalkers: we have egl features in screenshot, capture_target_frame_eglimage_on_display

Decodetalkers: and seems drm is not optional in this time.. also gbm

Gigas002: Sorry I was partially incorrect in wording. I mean we don't ever use this on `wayshot` side. It'd be better to separate this way though (`egl` should depend on `dmabuf`, not `screencast`):

```toml
dmabuf = ["dep:gbm", "dep:drm"]
egl = ["dep:gl", "dep:r-egl-wayland", "dmabuf"]
screencast = ["dmabuf"]
```

I have created a patch to confirm `wayshot` itself works good. We never use any `dmabuf`-related branches so these dependencies are not needed for our binary. Would be glad if you can check this out and share your thoughts.

[390_review.patch](https://github.com/user-attachments/files/30161950/390_review.patch)

Decodetalkers: can we apply this change in another pr? it is a little difficult to discussion in this thread..

Gigas002: Sure I can create a follow-up then

Decodetalkers: ok, it this pr is fine now, I want to merge it.. because  it was a serious bug..

Gigas002 (approved):
@Decodetalkers 
Thank you very much for the fix, LGTM!
I will create a follow-up PR for discussion on feature separations stuff

PR #392, opened 2026-07-15
feat: add nix module & repackage with feature flags

id3v1669:
What's done:
- add home manager module
- repackage to add support for build features with defaulting to one in repo
- move shell, package away from flake
- bump lock
- workflow with checks for nix eval, build & fmt

Why:
- toml is well supported format in nix, so nice to have a module
- move packaging and shell away from flake as it grew too big with module
- regarding features disabled when package is set - if user uses package from nixpkgs no features there present at least for now

ps. 
I thought about adding a regular module, not home manager. But it would require either changing wayshot code to support multiple `--config` flags and default to last one or create an ugly wrapper for that. That's why that idea on hold for now.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,21 @@
+name: Nix Checks
+
+on: [push, pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v31
+
+      - name: Check flake evaluation
+        run: nix flake check --no-build --all-systems
+
+      - name: Check formatting
+        run: nix fmt -- --check .
+
+      - name: Check build
+        run: nix build
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -1,0 +1,77 @@
+{
+  lib,
+  stdenv,
+  rustPlatform,
+  pkg-config,
+  installShellFiles,
+  pango,
+  libgbm,
+  libGL,
+  wayland,
+  jpegSupport ? true,
+  pnmSupport ? true,
+  qoiSupport ? true,
+  webpSupport ? true,
+  avifSupport ? true,
+  jxlSupport ? true,
+  clipboardSupport ? true,
+  colorPickerSupport ? true,
+  completionsSupport ? true,
+  loggerSupport ? true,
+  notificationsSupport ? true,
+  selectorSupport ? true,
+}:
+rustPlatform.buildRustPackage rec {
+  pname = "wayshot";
+  version = "${(builtins.fromTOML (builtins.readFile (src + "/Cargo.toml"))).workspace.package.version}-git";
+
+  src = lib.cleanSource ../.;
+
+  cargoLock.lockFile = "${src}/Cargo.lock";
+
+  buildNoDefaultFeatures = true;
+  buildFeatures =
+    lib.optional jpegSupport "jpeg"
+    ++ lib.optional pnmSupport "pnm"
+    ++ lib.optional qoiSupport "qoi"
+    ++ lib.optional webpSupport "webp"
+    ++ lib.optional avifSupport "avif"
+    ++ lib.optional jxlSupport "jxl"
+    ++ lib.optional clipboardSupport "clipboard"
+    ++ lib.optional colorPickerSupport "color_picker"
+    ++ lib.optional completionsSupport "completions"
+    ++ lib.optional loggerSupport "logger"
+    ++ lib.optional notificationsSupport "notifications"
+    ++ lib.optional selectorSupport "selector";
+
+  nativeBuildInputs = [
+    pkg-config
+    installShellFiles
+  ];
+
+  buildInputs = [
+    pango
+    libgbm
+    libGL
+    wayland
+  ];
+
+  postInstall = lib.optionalString (completionsSupport && stdenv.buildPlatform.canExecute stdenv.hostPlatform) ''
+    installManPage docs/wayshot.1.scd docs/wayshot.5.scd docs/wayshot.7.scd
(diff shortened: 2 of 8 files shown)


Diax170:
> I thought about adding a regular module, not home manager.

Maybe adding two modules would make sense, where the NixOS module would only have `enable`, `package` and `features` options and not `config`. For an example, see [Bezel modules](https://github.com/Indra55/bezel/tree/main/modules) which I PR'd in https://github.com/Indra55/bezel/pull/7

id3v1669:
> > I thought about adding a regular module, not home manager.
> 
> Maybe adding two modules would make sense, where the NixOS module would only have `enable`, `package` and `features` options and not `config`. For an example, see [Bezel modules](https://github.com/Indra55/bezel/tree/main/modules) which I PR'd in [Indra55/bezel#7](https://github.com/Indra55/bezel/pull/7)

In case of bezel it makes sense as there is a systemd service and uinput. In our case we have just package, so without config declaration there is no point in module as override does the thing.

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+      --bash <($out/bin/wayshot --completions bash) \
+      --fish <($out/bin/wayshot --completions fish) \
+      --zsh <($out/bin/wayshot --completions zsh)
+  '';
+
+  meta = {
+    description = "Screenshot crate for wlroots based compositors implementing the zwlr_screencopy_v1 protocol.";
+    homepage = "https://crates.io/crates/wayshot";
+    license = with lib.licenses; [
+      bsd2
```
Gigas002: The binary crate itself is `gpl3Only`, so I don't think `wayshot` users need `bsd2` here. It's for `libwayshot` crate users only. Kinda doesn't make much sense here IMO

nix/package.nix
```diff
@@ -0,0 +1,76 @@
...
+  ];
+
+  buildInputs = [
+    pango
+    libgbm
+    libGL
+    wayland
+  ];
+
+  postInstall = lib.optionalString (completionsSupport && stdenv.buildPlatform.canExecute stdenv.hostPlatform) ''
```
Gigas002, line 59: I think users should decide to create or not create completions depending on their own environment and they should not be created by default. We support [more shells](https://github.com/waycrate/wayshot/blob/d2b8ca737a43fea765edecbac8009f971d9be817/wayshot/src/cli.rs#L20), than just `bash`/`fish`/`zsh` for completions, so I'd say we either generate all or don't generate any by default.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,21 @@
+name: Nix Checks
+
+on: [push, pull_request]
```
Gigas002, line 3: I have mixed feelings about `pull_request` trigger. I feel like there's a possibility that a contributor does not care about nix build and doesn't know it's inner kitchen, but feature addition, e.g. new runtime dependency, breaks the job, and contributor will have to search for a way to fix it just in order to make CI pass for a system he doesn't care or know and didn't touch related files in PR in first place.
Addressing fixes on `main` branch's nix fails is a correct IMO.
Maybe doing this action dispatchable and dispatch it manually on nix-relaated PRs only is a good idea.
@Decodetalkers what's your thoughts?

Gigas002 (changes requested):
@id3v1669 
Thanks!
I think the job itself is a good idea since it will simplify nix-related contributions, but I'm not sure about the form. Would be happy if you can address the conversations.

.github/workflows/nix.yml
id3v1669: Maybe just not consider its failure as a blocker for pr, but an indicator to call me or you to update the nix stuff?
I mean even if not nix-related stuff is updated, but that breaks packaging we need a way to track that, so we could start preparing before it's merged.

nix/package.nix
id3v1669: About install by default, it's done that way in nixpkgs for all packages as far as I see. We can change that and make it a build flag.
Regarding shell support:
In [docs](https://github.com/NixOS/nixpkgs/blob/6bfaf02a46540dad5b083e46e80d9ac133260cd3/doc/hooks/installShellFiles.section.md?plain=1#L91) we can add nushell, but Elvish and Pwsh are not there.
I'll add nushell, but how do you want to proceed with install itself?
If you insist on making it optional, I would at least default to true to make life easier for regular users and follow nixpkgs.

Gigas002: Ah, if it's common for nixpkgs to have these then fine, I'd prefer to follow other packages's standard in that case. I'm not a nix user myself so feel free to correct me, I just thought it looked incorrect on the first sight.

Diax170:
> In our case we have just package, so without config declaration there is no point in module as override does the thing.

yeah that makes sense, I agree

Diax170:
maybe a basic question but do you put homeModules or homeManagerModules in flake.nix, or doesn't it matter?

Diax170:
maybe we could also PR similar changes to Nixpkgs (overlay) and Home Manager (hm module)?

id3v1669:
> homeManagerModules

I also thought `homeManagerModules` should be used.
It's valid for now, but `nix flake check --no-build --all-systems` throws warning:
`warning: unknown flake output 'homeManagerModules'`

Diax170:
> > homeManagerModules
> 
> I also thought `homeManagerModules` should be used. It's valid for now, but `nix flake check --no-build --all-systems` throws warning: `warning: unknown flake output 'homeManagerModules'`

yeah it's probably homeModules, sorry for confusion

.github/workflows/nix.yml
Gigas002: Ok, I think it's fine to not consider it as blocker for urgent PRs and call for help in other cases, I guess.

Gigas002 (approved):
@id3v1669 
Thank you, I hope it'll make easier to track `nix`-specific issues.

PR #393, opened 2026-07-16
feat: install manpages in Nix package

Diax170:
when adding shell completions in #386 I forgot to also add manpages, so here they are

will make a Nixpkgs PR soon

PR #394, opened 2026-07-17
fix: move manpages out of stdenv.hostPlatform.canExecute in flake.nix

Diax170:
While making #393 I accidentally put installManPage in the wrong scope, as addresed by @NickCao in https://github.com/NixOS/nixpkgs/pull/542651. This PR fixes that

@id3v1669 you'll probably need to update #392

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -23,7 +23,9 @@
 }:
 rustPlatform.buildRustPackage rec {
   pname = "wayshot";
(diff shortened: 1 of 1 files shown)


Diax170:
why did the build fail

id3v1669:
> why did the build fail

plz rebase, should be good now

Diax170:
> > why did the build fail
> 
> plz rebase, should be good now

thanks, it's fixed, you can review now

id3v1669 (approved):
Thanks!

PR #395, opened 2026-07-19
fix: separate dmabuf-related features on libwayshot side

Gigas002:
As discussed in: https://github.com/waycrate/wayshot/pull/390#discussion_r3603549139, separated features since we don't need to depend on that stuff in all scenarios and on `wayshot` binary side.

### DIFF
--- a/libwayshot/src/egl.rs
+++ b/libwayshot/src/egl.rs
@@ -124,6 +124,7 @@ pub fn bind_egl_image_to_gl_texture(guard: &EGLImageGuard) -> Result<()> {
 
 /// Create an EGLImage from the DMA-BUF and bind it to the current GL texture, then destroy the EGLImage.
 /// Used by screencast when updating the texture each frame.
+#[cfg(feature = "screencast")]
 pub fn create_egl_image_and_bind_to_gl_texture(
     egl_display: egl::Display,
     bo: &BufferObject<()>,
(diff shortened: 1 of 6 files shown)


libwayshot/src/egl.rs
```diff
@@ -124,6 +124,7 @@ pub fn bind_egl_image_to_gl_texture(guard: &EGLImageGuard) -> Result<()> {
 
 /// Create an EGLImage from the DMA-BUF and bind it to the current GL texture, then destroy the EGLImage.
 /// Used by screencast when updating the texture each frame.
+#[cfg(feature = "screencast")]
```
Decodetalkers, line 127: If it is the part of screencast,, can we just move it to screencast.rs?

Gigas002: We can, but that isn't a good idea IMO. The thing is, `create_egl_image_and_bind_to_gl_texture` is only consumed by `screencast.rs`, so without feature-gate we have dead code (`egl.rs` API is not public). 
We can either make `egl` a public API and remove `feature` attribute here, or leave it as-is. I don't want to put this code into `screencast` because it's in fact purely `egl` related code, and if we're to implement `vulkan` backend in future `screencast` will have backend-related code that it doesn't need to have.

Gigas002: or we can `#[allow(dead_code)]` here and remove `feature` statement

Decodetalkers: I see, then LGTM

Decodetalkers (approved):

Gigas002:
Thanks for review!

PR #396, opened 2026-07-20
fix: guards & build with --no-default-features --all-targets flags

id3v1669:
fix: guards & build with --no-default-features --all-targets flags


id3v1669:
Added workflow with --all-targets, so example would be checked too.
Nix workflow fixed as per [that](https://github.com/waycrate/wayshot/security/code-scanning/9)

Decodetalkers (approved):

Decodetalkers:
Thanks!

PR #398, opened 2026-07-20
chore(api)!: remove print_displays_info from libwayshot

Decodetalkers:
libwayshot should not do something like printing something, this logic should be moved to wayshot

as mentioned in https://github.com/waycrate/wayshot/pull/387#issuecomment-5020923062


Gigas002 (approved):
LGTM, thanks!

Shinyzenith (approved):

Decodetalkers:
Thank you so much!

PR #405, opened 2026-07-20
feat: allow screenshot_single_output with wl_output

Decodetalkers:
we cannot make a OutputInfo anytime, and with this api we do not need any other information

this will also help the waysip api


Decodetalkers:
I just was helping about the waysip api, and find this interface need a struct from libwayshot, so I change it a little

this will allow the freeze api can use wl_output to screenshot directly.

Gigas002 (approved):
LGTM, thank you!

PR #406, opened 2026-07-20
feat: add try_clone function for WayshotConnection

Decodetalkers:
this function will allow WayshotConnection to try to clone itself, keep most status, sharing the same wayland connection


Gigas002 (approved):
LGTM!

PR #407, opened 2026-07-20
refactor(screenshot_toplevel): remove Toplevel in Region and etc

Decodetalkers:
remove the Toplevel in enum Region since it is not the part of Region, and not use ToplevelInfo but with ExtToplevel to screenshot directly

### DIFF
--- a/libwayshot/src/lib.rs
+++ b/libwayshot/src/lib.rs
@@ -80,6 +80,7 @@ use wayland_protocols_wlr::{
 use crate::{
     dispatch::{CaptureFrameState, FrameState, OutputCaptureState, WayshotState},
     output::ToOutputInfo,
+    region::AsExtForeignToplevel,
 };
 
 pub use crate::{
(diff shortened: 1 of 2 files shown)


Gigas002:
@Decodetalkers 
Does the new generic bound actually add capability? I checked and `ExtForeignToplevelHandleV1` [doesn't implement](https://docs.rs/wayland-protocols/latest/wayland_protocols/ext/foreign_toplevel_list/v1/client/ext_foreign_toplevel_handle_v1/struct.ExtForeignToplevelHandleV1.html) `AsRef<Self>`, so a raw handle still can't be passed directly, only `TopLevel`/`&TopLevel` satisfy the bound today, same as before. 
If the intent is to eventually accept raw handles, that's fine to land.

Decodetalkers:
> @Decodetalkers Does the new generic bound actually add capability? I checked and `ExtForeignToplevelHandleV1` [doesn't implement](https://docs.rs/wayland-protocols/latest/wayland_protocols/ext/foreign_toplevel_list/v1/client/ext_foreign_toplevel_handle_v1/struct.ExtForeignToplevelHandleV1.html) `AsRef<Self>`, so a raw handle still can't be passed directly, only `TopLevel`/`&TopLevel` satisfy the bound today, same as before. If the intent is to eventually accept raw handles, that's fine to land.

Thank you for the check, now I add another trait for it

Gigas002 (approved):
LGTM thank you

Issue #410, opened 2026-07-22
rotation of outputs is ignored


Diax170:
If I rotate an output/mointor in sway using the `transform` option Wayshot ignores that and takes a screenshot in the output's default orientation. This is uncomfortable if someone uses portrait monitors

Decodetalkers:
> If I rotate an output/mointor in sway using the `transform` option Wayshot ignores that and takes a screenshot in the output's default orientation. This is uncomfortable if someone uses portrait monitors

If just one screen, I think just rotate it yourself will be better. So it is ignored. If it crosses the screen , then it will be rotated.

Decodetalkers:
Ok, here

https://github.com/waycrate/wayshot/blob/main/libwayshot%2Fsrc%2Fscreencopy.rs#L88-L92

We should use ImageBuffer from_fn instead, to apply the transformer directly, during it was saved. I will finish it later

Decodetalkers:
fixed. I will first publish a rc version for it..

PR #411, opened 2026-07-23
refactor: better rotation logic

Decodetalkers:
not use to_vec anymore
the rotation is fixed after FrameCopy is converted to Image. delete ImageBuffer<Mmap>
delete the functions relate to the rotation and resize


Decodetalkers:
this will also fix https://github.com/waycrate/wayshot/issues/410

and after this fix, I think the freeze option of waysip will also become normal after rotation

Gigas002:
The resulting screenshot is good, but `freeze` overlay is upside-down for me?
Side note: `waysip -d --freeze` requires separate fix in it's repo

Decodetalkers:
> The resulting screenshot is good, but `freeze` overlay is upside-down for me?
> Side note: `waysip -d --freeze` requires separate fix in it's repo

You mean the direction of the image saved is right but the image used by waysip is in wrong direction? Which rotation has that problem? I think I have tested them all, with this patch for waysip

Gigas002:
The waysip's `freeze` flag is different implementation from `libwayshot`'s freeze, so it's a different issue.
I'm talking about `wayshot -g` here, about rotation issue. Maybe different `transform` value is not being processed properly, idk?
E.g. [hyprland](https://wiki.hypr.land/Configuring/Basics/Monitors/#rotating) has 7 transform ways for output:

```
0 -> normal (no transforms)
1 -> 90 degrees
2 -> 180 degrees
3 -> 270 degrees
4 -> flipped
5 -> flipped + 90 degrees
6 -> flipped + 180 degrees
7 -> flipped + 270 degrees
```

In my case, I use `3`, and the vertical output is upside down. The image itself is not distorted, and when I select the area via geometry selector resulting output is correct. Only frozen screen, preview, is affected.

Decodetalkers:
> The waysip's `freeze` flag is different implementation from `libwayshot`'s freeze, so it's a different issue. I'm talking about `wayshot -g` here, about rotation issue. Maybe different `transform` value is not being processed properly, idk? E.g. [hyprland](https://wiki.hypr.land/Configuring/Basics/Monitors/#rotating) has 7 transform ways for output:
> 
> ```
> 0 -> normal (no transforms)
> 1 -> 90 degrees
> 2 -> 180 degrees
> 3 -> 270 degrees
> 4 -> flipped
> 5 -> flipped + 90 degrees
> 6 -> flipped + 180 degrees
> 7 -> flipped + 270 degrees
> ```
> 
> In my case, I use `3`, and the vertical output is upside down. The image itself is not distorted, and when I select the area via geometry selector resulting output is correct. Only frozen screen, preview, is affected.

Cannot reproduce the problem..

Decodetalkers:
> The waysip's `freeze` flag is different implementation from `libwayshot`'s freeze, so it's a different issue. I'm talking about `wayshot -g` here, about rotation issue. Maybe different `transform` value is not being processed properly, idk? E.g. [hyprland](https://wiki.hypr.land/Configuring/Basics/Monitors/#rotating) has 7 transform ways for output:
> 
> ```
> 0 -> normal (no transforms)
> 1 -> 90 degrees
> 2 -> 180 degrees
> 3 -> 270 degrees
> 4 -> flipped
> 5 -> flipped + 90 degrees
> 6 -> flipped + 180 degrees
> 7 -> flipped + 270 degrees
> ```
> 
> In my case, I use `3`, and the vertical output is upside down. The image itself is not distorted, and when I select the area via geometry selector resulting output is correct. Only frozen screen, preview, is affected.

I see, this problem maybe do not related with my change. the freeze feature just use the buffer by the protocol, and I did not touch the logic about this part, what I changed is the part of saving. I think the problem also exists before this pr

https://github.com/waycrate/wayshot/blob/main/libwayshot/src/lib.rs#L1313-L1358

Decodetalkers:
> The waysip's `freeze` flag is different implementation from `libwayshot`'s freeze, so it's a different issue. I'm talking about `wayshot -g` here, about rotation issue. Maybe different `transform` value is not being processed properly, idk? E.g. [hyprland](https://wiki.hypr.land/Configuring/Basics/Monitors/#rotating) has 7 transform ways for output:
> 
> ```
> 0 -> normal (no transforms)
> 1 -> 90 degrees
> 2 -> 180 degrees
> 3 -> 270 degrees
> 4 -> flipped
> 5 -> flipped + 90 degrees
> 6 -> flipped + 180 degrees
> 7 -> flipped + 270 degrees
> ```
> 
> In my case, I use `3`, and the vertical output is upside down. The image itself is not distorted, and when I select the area via geometry selector resulting output is correct. Only frozen screen, preview, is affected.

Maybe we should report it to hyperland.. Maybe there are something wrong with their layershell implement.. this bug has nothing to do with this pr.. 

I will merge this pr. If there is any problem, we can still continue to fix it. I think this fix is right, and will fix many problems

Gigas002:
Filed the report to hyprland, thanks for looking into it as well

PR #412, opened 2026-07-23
chore(tidy up): make the name style the same, remove wrong use unwrap

Decodetalkers:
make the name style of the api to same, and should not unwrap the protocol, should check them and return error if they are none


Gigas002 (approved):
LGTM!

PR #418, opened 2026-07-27
fix: allow negative coordinates in geometry string

### LINKED ISSUE #413: bug: providing out-of-bounds geometry string results in panic/incorrect screenshot
Providing out-of-bounds values to wayshot's `-g` results in panic and incorrect screenshots:

In case of negative value, it's option parsing error:

```sh
❯ wayshot -g "-498,61 648x475" | wl-copy
error: unexpected argument '-4' found

  tip: to pass '-4' as a value, use '-- -4'

Usage: wayshot [OPTIONS] [FILE]

For more information, try '--help'.
```

In case of out-of-bounds content it's black nothingness:

https://github.com/user-attachments/assets/7bc3c315-fa71-44cc-91e3-ce32274227b2

Resulting screenshot: 

Image

theripper-1920:
## Summary

Allow hyphen prefixed values for the `--geometry` argument so that negative coordinates are parsed correctly.

Previously, commands such as:

```bash
wayshot -g "-100,-100 200x200"
```

would fail to parse due to the leading hyphen being treated as an unknown CLI flag. This fix allows signed integers in geometry strings, which is required for multi-monitor setups (e.g., in Hyprland) where displays are positioned to the left or top of the primary monitor.

Fixes #413

### DIFF
--- a/wayshot/src/cli.rs
+++ b/wayshot/src/cli.rs
@@ -104,7 +104,7 @@ pub struct Cli {
     /// Without a value: interactively select a region (requires the `selector` feature).
     /// With a value: parse the geometry string in slurp/grim format "x,y widthxheight".
     ///   Example: wayshot -g "$(slurp)" or wayshot -g "783,746 177x251"
-    #[arg(short, long, value_name = "GEOMETRY", num_args = 0..=1, verbatim_doc_comment)]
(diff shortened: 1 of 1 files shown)


Gigas002 (approved):
Great, thanks!

elde-n:
Caused a slight issue for me with the geometry flag allowing `None` for the custom selector, had to switch the command from `wayshot -g --clipboard --silent -` to `wayshot --clipboard --silent - -g`

PR #424, opened 2026-08-04
feat: always save to stdout if it is in pipe

Decodetalkers:
this will allow save always work if there is a pipe operator exist.

### DIFF
--- a/wayshot/src/wayshot.rs
+++ b/wayshot/src/wayshot.rs
@@ -1,4 +1,4 @@
-use std::io::{self, BufWriter, Write};
+use std::io::{self, BufWriter, IsTerminal, Write};
 use std::time::Duration;
 
 use clap::Parser;
@@ -153,7 +153,7 @@ fn main() -> Result<()> {
                         std::fs::write(f, &encoded)?;
                     }
 
-                    if settings.stdout_print {
+                    if settings.stdout_print || !stdout.is_terminal() {
                         writer.write_all(&encoded)?;
                     }
 


Decodetalkers:
I just want if this pr can be included in 1.6.0 @Gigas002 . I just know we can check if the program is in pipe, then we can decide whether to write to stdout or not. I think it will be quite helpful for people who use the terminal.

Gigas002:
Honestly, I'm against merging it as-is...

We already have two explicit ways to get stdout output: pass `-` as the FILE arg, or set `stdout = true` in the config. Both are opt-in and predictable. This PR replaces that explicit contract with an implicit one based on `is_terminal()`, which means *any* non-interactive stdout — not just an intentional pipe to another image tool — now gets raw image bytes written to it.

That breaks a legitimate use case: someone redirecting wayshot's stdout for logging/debugging (e.g. `wayshot >> ~/wayshot.log 2>&1` in a keybinding wrapper, or output captured by a supervisor/systemd) will suddenly get binary image data mixed into what they expected to be plain log output. There's no flag to opt out of that once it's the default.

Decodetalkers:
ok.. let me think about it later

Decodetalkers:
But in fact, our log is written to stderr, it won't enter stdout

Decodetalkers:
I am just thinking about this case.. I want it always save to file when I do not use the pipe command, and this time it should not write anything to the stdout. 

And when I am using pipe command, I want the stdout contains the image data.. 

And since the log is in stderr, so I think out of this pr

Gigas002:
Hmm, I see. I think you are correct about logs.
Still, I personally would like to not introduce breaking behavior that cannot be turned on/off. We can change the config's resolvement to always send screenshot to stdout:

```toml
[base]
stdout = true
```

This would affect all `wayshot` usecases, but won't introduce unchangable behavior, users will be able to disable it back to `false`.
Or we can add a new flag for this, but having different flag for something we already have sounds a bit like overkill to me...

Decodetalkers:
ok, let we discus it later. Do not let it in 1.6.0 now

Decodetalkers:
And I have noticed a problem.. if next time there is only something change in wayshot.. will cargo publish --workspace work as normal? ..

Decodetalkers:
if we want to induce this behavior. Then we may have 4 behavior

Save to file, not save to stdout
Not save to file, always to stdout
Save to file, and save to stdout
behavior will be decided by that if pipe is used

PR #425, opened 2026-08-04
fix: prefer the transform by ext-image-copy

### LINKED ISSUE #359: Region Aware Multimonitor Support
`wayshot -g` seems to fail to accurately capture the selected region on a multi-monitor system. Parts of the second monitor seem to be overlaid on top of the selection.

Decodetalkers:
resolve: https://github.com/waycrate/wayshot/issues/359

before we use the transform by wl-output to address the final image, but it is wrong. now ext-image-copy provide us the real transform of the image, we need to use it first, if we have to use the wlr-screencopy, then we use the transform of the screen


Gigas002 (approved):
LGTM

PR #426, opened 2026-08-05
chore: proper test coverage

### LINKED ISSUE #408: enhancement: improve test coverage
We need to investigate which areas are not yet covered and could be covered better.
Also, tests for `wayshot` crate are needed. The architecture must follow existing `libwayshot` pattern: test must coexist in separate directory.
The priority isn't very high, but I'd like this to be implemented before next release. After this change, all further contributions will have to add tests for new implemented code.
Currently, it's up for grabs, so anyone willing to do this can comment and start out with implementation, just make sure to communicate please.

Gigas002:
Used this script to track the uncovered areas and check them out in browser as `html`:

```sh
export LLVM_COV=/usr/bin/llvm-cov LLVM_PROFDATA=/usr/bin/llvm-profdata
cargo llvm-cov --all-features --workspace --html --output-dir target/llvm-cov
```

Closes: #408


Gigas002:
This PR also fixes one --clipboard bug and adds headless sway in CI for real compositor testing. Some bits, like `egl` are not testable withot GPU, and some require interactive selection, like `color-picker`, so there's nothing we can do, at least for CI tests.
Coverage is up from 8% to 64%

Adding headless compositor capabilities to CI may also help introducing automated benchmarks in future...

Gigas002:
Or 70% depending on how to count

image

wayshot/src/clipboard.rs
```diff
@@ -47,8 +47,14 @@ pub fn copy_to_clipboard(data: Vec<u8>, encoding: EncodingFormat) -> Result<()>
...
         Ok(Fork::Child(_)) => {
+            if let Ok(devnull) = std::fs::File::open("/dev/null") {
+                let _ = rustix::stdio::dup2_stdin(&devnull);
+                let _ = rustix::stdio::dup2_stdout(&devnull);
+                let _ = rustix::stdio::dup2_stderr(&devnull);
+            }
             opts.foreground(true);
-            opts.copy(Source::Bytes(data.into()), mime)?;
+            let result = opts.copy(Source::Bytes(data.into()), mime);
+            std::process::exit(if result.is_ok() { 0 } else { 1 });
```
Decodetalkers, line 57: what is this change for? Will it solve the problem that clipboard event won't go well?

wayshot/tests/cli_integration.rs
```diff
@@ -0,0 +1,367 @@
...
+//! connect. None of this requires a running compositor, so it's safe in
+//! headless CI. `WAYLAND_DISPLAY`/`WAYLAND_SOCKET` are stripped explicitly so
+//! the tests behave the same on a machine that does have a compositor
+//! running.
+
+use std::process::Command;
+
+fn wayshot_cmd() -> Command {
+    let mut cmd = Command::new(env!("CARGO_BIN_EXE_wayshot"));
+    cmd.env_remove("WAYLAND_DISPLAY");
```
Decodetalkers, line 16: Is this logic to prevent wayshot from running? so the test does not contain the wayland logic? I want to confirm that

wayshot/src/clipboard.rs
Gigas002: I discovered this bug when was creating tests. When you run:

```sh
wayshot --clipboard - | wl-copy
```

after the clipboard event ended, it had ran the notification part once more for the fork child. Now it exists properly.

Decodetalkers:
I just wonder if one day archlinux package our package, how will they run the unit test.. they will install sway for too many times www

Decodetalkers:
It should be a good way to test these features.. but will these unit test pass be skipped when there is not a sway wm for it to test?

Oh.. I see, seems it is fine.. I will read this pr carefully after I come back home

wayshot/tests/cli_integration.rs
Gigas002: Yes, this is correct. These tests only the CLI and main() logic, not the Wayland functionality itself

Gigas002:
I think I can add a configuration flag for tests that require `sway` so that downstream won't have to bother with them, running only unit tests... Though I doubt downstream even need to run the tests, it's our job as developers of software

Decodetalkers:
> I think I can add a configuration flag for tests that require `sway` so that downstream won't have to bother with them, running only unit tests... Though I doubt downstream even need to run the tests, it's our job as developers of software

They will. so I just a little worry about it . But we can just make it pass when there is no display

Decodetalkers:
We should use a feature to hide those test that only will pass in ci environment

libwayshot/src/tests/live_capture.rs
```diff
@@ -0,0 +1,111 @@
...
+    let image = conn
+        .screenshot_freeze(move |_conn| Ok(expected_region), false)
+        .expect("freeze capture with a programmatic callback should succeed");
+    assert!(image.width() > 0);
+    assert!(image.height() > 0);
+}
+
+#[cfg(feature = "dmabuf")]
+#[test]
+fn dmabuf_capture_errors_cleanly_without_a_dmabuf_capable_connection() {
```
Decodetalkers, line 97: this will only work in ci, because in ci, there is no dmabuf support

Decodetalkers:
Ok, only find one problem. I think most can be passed. Thank you very much!

libwayshot/src/tests/live_capture.rs
Decodetalkers: if one with the real environment run this test, it won't pass, so it needs to be hidden behind feature

Gigas002: Whoops, sorry, the comment there was incorrect. The test is actually fine and doesn't depend on the CI environment. This test runs fine on my machine.
Rewrote the comment to explain that instead of implying it's a CI-only thing.

Decodetalkers: I see, well done! let we ship this pr

Decodetalkers (approved):
LGTM

PR #430, opened 2026-08-08
fix: make man pages actualy detectable

id3v1669:
update packaging & fix man pages detection.

ps. don't use man pages, thanks to my mate for noticing the issue


Gigas002 (approved):
Thanks!

